Issue #10, opened 2015-06-03
Add an optional tag to define required input file extensions (potentially multiple extensions).


glatard:
or regexp on the file name.

Issue #4, opened 2015-05-14
Add optional inputs and outputs


glatard:

glatard:
Think of mutually exclusive inputs too.

glatard:
Think of optional outputs too.

Issue #2, opened 2015-05-14
Add flags


glatard:

glatard:
Adding a "Flag" type seems sufficient. A flag has an command-line key, which must contain the value of the flag (e.g. "-T"). At runtime, the command-line key remains untouched when the flag is true, and it is removed from the command line when the flag is false.

PR #12, opened 2015-06-14
#4 optional inputs and outputs; #2 Flag input type; #10 input file extensions.

glatard:
Created two new optional attributes for input and output:
* optional is a boolean (False if not present) that describes if the input/output is optional.
* command-line-flag is a string that defines the flag to prepend to the input/output value of any input. This is necessary to support optional inputs and outputs as their command-line flag shouldn't be put on the command-line if the input/output is not present. 
* optional outputs are output files that may or may not be produced by the tool


glatard:
Created a Flag input type. 
* A Flag must have a command-line-flag.
* When a Flag input is activated, the command-line-key is replaced by the command-line-flag.

Issue #58, opened 2016-09-12
Support for more containers types


glatard:
A tool may contain one and only one of the following attributes:
- Docker index (e.g. `http://index.docker.io`) and image name (e.g. `mcin/ica-aroma:latest`). This is already possible in the schema.
- URL of a rootfs container (e.g. `http://somehwere.org/container-1234.tgz`).
- URL of a singularity container (e.g. `http://somehwere.org/container-singularity-1234.tgz`).

An optional attribute containing the hash of the container may be present. 

A parent `container` object should be created to encapsulate these different implementations.

ttaa9:
Maybe have a `containers` array, which could reference multiple images/containers?

glatard:
No because these implementations must be exclusive: only one can be specified.

kofalt:
SGTM. Since each option might have multiple values, what would you think of something like this?

```js
{
	// ... other boutique values ...

	"input": {
		"type": "docker",
		"index": "http://index.docker.io",
		"image": "mcin/ica-aroma:latest"
	}
}
```

```js
{
	"input": {
		"type": "rootfs",
		"uri": "http://somehwere.org/container-1234.tgz"
	}
}
```

This way, you could use [definitions](https://spacetelescope.github.io/understanding-json-schema/structuring.html#reuse) and the [oneOf](https://spacetelescope.github.io/understanding-json-schema/reference/combining.html#oneof) keyword to formally validate each type.

chrisfilo:
Minor comment: singularity does not yet support .tgz - the images use .img extension

vsoch:
Apologies for being the annoying person to comment on a closed issue - Singularity images support [multiple formats](https://github.com/singularityware/singularity/blob/ac48c7beab7c2b97c71b06e9ad6ab84400ef9dbb/libexec/cli/import.help#L22) and one of them is indeed .tgz. A more complete list of the "weird" ones:

![image](https://cloud.githubusercontent.com/assets/814322/21474861/c21a36c4-cad9-11e6-96e5-84c98332f954.png)

Issue #51, opened 2016-08-15
Work and file paths


glatard:
Some applications:
- need to be launched from a specific work directory in a container (add a string property for this). 
- only accept absolute file paths (add a boolean property to File inputs and outputs). 
- only accept relative file paths (add a boolean property to File inputs and outputs).

glatard:
Note: input and output relative paths must be defined from the work directory.

ttaa9:
I can understand this makes sense, in that it allows for greater customization.

But for cbrain specifically, I am not clear how this should be implemented. 

- For non-containerized tasks, the working directory is dynamically created based on the task id. So I don't know what specifying a launch directory would do. I'm also not sure how specifying absolute paths would be useful (for this non-docker case in cbrain). 

- For cbrain processes inside docker containers, would specifying the launch-directory simply mean we mount pwd to the target launch-directory, and then set that directory to the pwd (with `docker run ... -w`)? So the launch-directory must be an absolute path in this case, right? 

- What exactly would specifying that a file output path is absolute or relative do in cbrain? Should it check the output path value for each file and fail to launch if it is not absolute/relative? Also, for file inputs, they are given as userfiles and then their relative name only is written into the command template, I believe. So if `absolute-path` is given as true, should the input instead be replaced by the absolute path (and work as normal in all other cases)?

glatard:
Yes, some applications only work from a particular directory.

* This is only for the containerized tasks. Should be part of the `container` object mentioned in #58.

* For CBRAIN, it won't change anything to the mounts. Mounts have to remain as is, otherwise symbolic links will be completely broken. We would only pass an extra `-w` option and possibly change file paths.

* For input files: if a path relative to the cbrain task execution dir is passed, then it should be transformed into a path relative to the container workdir. If an absolute path is passed, then nothing has to be changed.

* For output files: if the application produces files outside of the cbrain execution directory, they should be copied in the cbrain execution directory because this is the only one which is mounted. 

In fact I'm not sure that we really want to go for the container workdir, as it will complexify file management quite a lot and may lead to tricky bugs...

ttaa9:
I see, so we would mount the same `-v ${PWD}:${PWD}` from the dynamically generated gridshare directory, but pass `-w container_workdir`. And then redirect all the relative input paths to be relative to the new pwd, and copy all the outputs back to the original mounted folder afterwards.

[...]

PR #61, opened 2016-09-19
Add path constraints to Files, and support for more containers.

ttaa9:
Attempts to satisfy the requirements of #51 and #58. 

If the schema changes seem ok, let me know and I will update the tools (and cbrain) as well.

### DIFF
--- a/schema/descriptor.schema.json
+++ b/schema/descriptor.schema.json
@@ -28,10 +28,9 @@
             "description": "A string that describes the tool command line, where input and output values are identified by \"keys\". At runtime, command-line keys are substituted with flags and values.",
             "type": "string"
         },
-        "container": {
-            "id": "http://github.com/boutiques/boutiques-schema/container",
+        "container-image": {
+            "id": "http://github.com/boutiques/boutiques-schema/container-image",
             "type": "object",
-            "description": "Object representing a container for the tool.",
             "allOf": [{
                 "properties": {
                     "working-directory": {
@@ -65,7 +64,8 @@
                             "type": "string"
                         }
                     },
-                    "required": ["type", "image"]
+                    "required": ["type", "image"],
+                    "description": "Object representing a docker container for the tool."
                 }, {
                     "properties": {
                         "type": { "enum": ["rootfs", "singularity"] },
@@ -76,7 +76,8 @@
                             "type": "string"
                         }
                     },
-                    "required": ["type", "url"]
+                    "required": ["type", "url"],
+                    "description": "Object representing a rootfs or singularity container for the tool."
                 }]
             }]
         },
@@ -297,14 +298,9 @@
                         "description": "Specify the maximum number of entries in the list. May only be used with List type inputs.",
                         "type": "number"
                     },
-                    "absolute-path": {
-                        "id": "http://github.com/boutiques/boutiques-schema/input/absolute-path",
-                        "description": "Specifies that this input must be given as an absolute path. Only specifiable for File type inputs. Cannot be true if relative-path is true.",
-                        "type": "boolean"
-                    },
-                    "relative-path": {
-                        "id": "http://github.com/boutiques/boutiques-schema/input/relative-path",
-                        "description": "Specifies that this input must be given as a relative path, with respect to the working directory. Only specifiable for File type inputs. Cannot be true if absolute-path is true.",
+                    "uses-absolute-path": {
+                        "id": "http://github.com/boutiques/boutiques-schema/input/uses-absolute-path",
+                        "description": "Specifies that this input must be given as an absolute path. Only specifiable for File type inputs.",
                         "type": "boolean"
                     }
                 },
@@ -313,34 +309,16 @@
                     "id",
                     "type"
                 ],
-                "allOf": [
-                    {
-                        "anyOf": [{
-                            "properties": {
-                              "type": { "enum": ["Flag"] },
-                              "list": { "enum": [false] }
-                            }
-                        }, {
-                            "properties": {
-                              "type": { "enum": ["String", "File", "Number", "Enum"] }
-                            }
-                          }
-                        ]
-                    },
-                    {
-                        "anyOf": [{
-                            "properties": {
-                                "absolute-path": { "enum": [true] },
-                                "relative-path": { "not": { "enum": [true] } }
-                            }
-                        }, {
-                            "properties": {
-                                "relative-path": { "enum": [true] },
-                                "absolute-path": { "not": { "enum": [true] } }
-                            }
-                        }]
+                "anyOf": [{
+                    "properties": {
+                      "type": { "enum": ["Flag"] },
+                      "list": { "enum": [false] }
                     }
-                ],
+                }, {
+                    "properties": {
+                      "type": { "enum": ["String", "File", "Number", "Enum"] }
+                    }
+                }],
                 "additionalProperties": true,
                 "dependencies": {
                     "command-line-flag-separator": ["command-line-flag"],
@@ -376,12 +354,7 @@
                             "type": {"enum": ["Number"]}
                         }
                     },
-                    "absolute-path": {
-                        "properties": {
-                            "type": {"enum": ["File"]}
-                        }
-                    },
-                    "relative-path": {
+                    "uses-absolute-path": {
                         "properties": {
                             "type": {"enum": ["File"]}
                         }
--- a/schema/descriptor.schema.json
+++ b/schema/descriptor.schema.json
@@ -28,18 +28,57 @@
             "description": "A string that describes the tool command line, where input and output values are identified by \"keys\". At runtime, command-line keys are substituted with flags and values.",
             "type": "string"
         },
-        "docker-image": {
-            "id": "http://github.com/boutiques/boutiques-schema/docker-image",
-            "minLength": 1,
-            "description": "Name of a Docker image where tool is installed and configured. Example: docker.io/neurodebian.",
-            "type": "string"
-        },
-        "docker-index": {
-            "id": "http://github.com/boutiques/boutiques-schema/docker-index",
-            "minLength": 1,
-            "description": "Docker index where Docker image is available.",
-            "default": "http://index.docker.io",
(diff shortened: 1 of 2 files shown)


schema/descriptor.schema.json
```diff
@@ -28,18 +28,57 @@
...
+                        },
+                        "index": {
+                            "id": "http://github.com/boutiques/boutiques-schema/container/index",
+                            "minLength": 1,
+                            "description": "Index where the image is available.",
+                            "default": "http://index.docker.io",
+                            "type": "string"
+                        }
+                    },
+                    "required": ["type", "image", "index"]
```
glatard: I'm not sure if `index` should be required since there is a default value for it.

schema/descriptor.schema.json
```diff
@@ -28,18 +28,57 @@
...
-            "id": "http://github.com/boutiques/boutiques-schema/docker-index",
-            "minLength": 1,
-            "description": "Docker index where Docker image is available.",
-            "default": "http://index.docker.io",
-            "type": "string"
+        "container": {
+            "id": "http://github.com/boutiques/boutiques-schema/container",
+            "type": "object",
+            "description": "Object representing a container for the tool.",
+            "allOf": [{
```
glatard: I don't understand why `allOf` is required here.

schema/descriptor.schema.json
```diff
@@ -28,18 +28,57 @@
...
+                "properties": {
+                    "launch-directory": {
```
glatard: I would call it `working-directory` to be consistent with the terminology used in `Docker`.

schema/descriptor.schema.json
ttaa9: It allows potentially letting `container-hash` and `launch-directory` coexist with either the `docker` spec or the `rootfs/singularity` spec. It does not force them to be required. It just lets both possible specs match to these so that the same values dont have to be written out twice (i.e. once in each of the two container type specs).

schema/descriptor.schema.json
ttaa9: Fair enough

schema/descriptor.schema.json
ttaa9: Sure

schema/descriptor.schema.json
glatard: OK

glatard: Well, isn't it possible to have a container object without an implementation specified then? For instance an empty object?

glatard: -> it's not possible thanks to the `required` property in *both* docker and rootfs/singularity sections, right?

ttaa9: That's correct; it has to satisfy exactly one of (`oneOf`) those two schemas, and each has its own `required` properties. Testing with an empty container gives an error because of this.

glatard:
@kofalt @chrisfilo @johnflavin would you like to review this PR too?

kofalt:
@glatard Totally! I'll see if I can give it a look today, but it might end up being tomorrow morning :smile:

chrisfilo:
I'm not trying to be picky, but isn't "image" a more suitable name than "container"?

schema/descriptor.schema.json
```diff
@@ -257,25 +296,50 @@
...
                         "description": "Specify the maximum number of entries in the list. May only be used with List type inputs.",
                         "type": "number"
+                    },
+                    "absolute-path": {
+                        "id": "http://github.com/boutiques/boutiques-schema/input/absolute-path",
+                        "description": "Specifies that this input must be given as an absolute path. Only specifiable for File type inputs. Cannot be true if relative-path is true.",
+                        "type": "boolean"
+                    },
+                    "relative-path": {
+                        "id": "http://github.com/boutiques/boutiques-schema/input/relative-path",
```
johnflavin: Is it necessary to have two mutually-exclusive boolean properties? If you set one to `true`, the other must be `false`. So maybe there could just be one input.

On the other hand, there is the middle case where the image doesn't care about whether the paths are absolute or relative, so neither property would be true. That is a benefit to having two properties. But in that case the tool implementation would have to decide at launch time to either make the path relative or absolute; it wouldn’t matter which one but it does have to set one or the other. But we could guide the tool writers by setting one of the two properties `true` by default.

I guess what I would want is for there to be one boolean property. If `true`, then the paths must be absolute; if `false`, then relative. (Or the other way. Whatever.) And one of those (absolute?) should be the default value if nothing is specified. No guess work by the tool developer.

tools/validator.rb
```diff
@@ -49,10 +49,11 @@ def usage
 ### Validation of descriptor arguments ###
 
 ## Helper functions ##
-inputGet  = lambda { |s| descriptor['inputs'].map {       |v| v[s] } rescue [] }
-outputGet = lambda { |s| descriptor['output-files'].map { |v| v[s] } rescue [] }
-groupGet  = lambda { |s| descriptor['groups'].map {       |v| v[s] } rescue [] }
-inById    = lambda { |i| descriptor['inputs'].find{       |v| v['id']==i } || {} }
+safeGet   = lambda { |sec,targ| descriptor[sec].map { |v| v[targ] }.compact rescue [] }
```
johnflavin: Is this stuff in scope of this PR?

glatard:
@chrisfilo you're right, how about `container-image`?

tools/validator.rb
glatard: Yes you're right, small fix though, let's keep it here if you don't mind.

chrisfilo:
+1

schema/descriptor.schema.json
glatard: The idea was that if none of these properties are here, then it is up to the implementation to decide which one to use. There might be good reasons to use one or the other. In CBRAIN for instance, relative paths are always preferred because it makes the tasks relocatable in different clusters (which is not really an issue with containers but still is the preferred way). 

But I think I'm fine with a single `use-absolute-path` boolean property, the default being to use relative paths.

schema/descriptor.schema.json
```diff
@@ -28,18 +28,57 @@
...
+                        "type": "string"
+                    },
+                    "container-hash": {
+                        "id": "http://github.com/boutiques/boutiques-schema/container/container-hash",
+                        "minLength": 1,
+                        "description": "Hash for the given container.",
+                        "type": "string"
+                    }
+                }
+            },{
```
kofalt: Related to the earlier conversation: couldn't you move this `properties` key up to the `container` level? Then you could ditch the `allOf` -> `oneOf` nesting (you'd just need `oneOf`).

schema/descriptor.schema.json
```diff
@@ -28,18 +28,57 @@
...
+                        "description": "Location from which this task must be launched within the container.",
+                        "type": "string"
+                    },
+                    "container-hash": {
+                        "id": "http://github.com/boutiques/boutiques-schema/container/container-hash",
+                        "minLength": 1,
+                        "description": "Hash for the given container.",
+                        "type": "string"
+                    }
+                }
```
kofalt: Related to the earlier conversation: couldn't you move this `properties` key up to the `container` level? Then you could ditch the `allOf` -> `oneOf` nesting (you'd just need `oneOf`).

schema/descriptor.schema.json
```diff
@@ -28,18 +28,57 @@
...
+                        },
+                        "index": {
+                            "id": "http://github.com/boutiques/boutiques-schema/container/index",
+                            "minLength": 1,
+                            "description": "Index where the image is available.",
+                            "default": "http://index.docker.io",
+                            "type": "string"
+                        }
+                    },
+                    "required": ["type", "image"]
```
kofalt: Add a `description` key here talking about what kind of container this is (docker)

schema/descriptor.schema.json
```diff
@@ -28,18 +28,57 @@
...
+                    "properties": {
+                        "type": { "enum": ["rootfs", "singularity"] },
+                        "url": {
+                            "id": "http://github.com/boutiques/boutiques-schema/container/url",
+                            "minLength": 1,
+                            "description": "URL where the image is available.",
+                            "type": "string"
+                        }
+                    },
+                    "required": ["type", "url"]
```
kofalt: Add a `description` key here talking about what kind of container this is (singularity or rootfs)

kofalt:
Review complete.

I'm not a fan of the absolute / relative feature here - I'm wondering if we can simplify things a bit?

I know of tools that require an absolute path, but I don't know of tools that *require* a relative path. Do you think we could get away with having just a boolean such as `absolutePathRequired`? If true, path must be absolute, if false, anything goes?

Motivation is that, like @glatard's comment on #51, I am thinking about the complexity of the command-generation bit and it would be nice to keep that simpler wherever possible. There's also something to be said for keeping the schema from getting too complex...

schema/descriptor.schema.json
ttaa9: I prefer a single `use-absolute-path` as well. 

One minor thing is the semantics of this property: does it mean that an input must be absolute (i.e. it just checks it) vs the platform must transform it, or is the interpretation up to the platform?

schema/descriptor.schema.json
ttaa9: I greatly like that idea, but I don't immediately see a clean way to do it, because (with the current names) the `required` property depends on which `oneOf` is being matched, and `required` must be on the same level as `properties`. (If you know of one, please let me know!)

schema/descriptor.schema.json
johnflavin: My interpretation is that if an input specifies `use-absolute-path=true`, then the platform must give that input an absolute-path value before it attempts to launch the container. The contract implied by this property is between the platform and the container. 

The contract between platform and user is another matter, and not covered by this property. The platform could require the user to give absolute paths to inputs with this property, or could translate user-provided relative paths to absolute. Whichever.

schema/descriptor.schema.json
glatard: @kofalt any suggestion on this?

PR #66, opened 2016-10-18
Documents the latest schema additions.

glatard:
@tta9: please check that all your latest additions are now documented.


kofalt:
LGTM!

Might be too late in the game, but what would you guys think about corralling custom properties under a common key? This groups similar information together and makes it clear that it's an extension.

We did this on our side, making `custom` an optional key that must be an object. Then we can set `additionalProperties` to false and have a cleaner base.

glatard:
Not too late @kofalt, see #72.

ttaa9:
I've always wanted to set `additionalProperties` to false, especially for detecting name bugs (e.g. very easy to write `disables` rather than `disables-inputs`). 👍

kofalt:
Awesome. Sounds like we can tackle that after this PR then :)

ttaa9:
@glatard Looks good. I'll add a little section on the invocationSchemaHandler in my next PR.

PR #73, opened 2016-10-19
Path handling

ttaa9:
Updates localExec.py to the latest schema version.

Includes support for `uses-absolute-paths` and specifying a `working-directory` for containers in the boutiques descriptor. Also added the option to run containers as the current user (so that output files are not owned by root).

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -23,7 +23,7 @@ Note: there is a Docker image ([boutiques/boutiques](https://hub.docker.com/r/bo
      descriptor and validating input data examples with respect to these invocation schemas.
   * Usage examples:
     * Generates an invocation schema: `python invocationSchemaHandler.py -i toolDescriptor.json`
-    * Validate input data (i.e. tool parameter arguments) with implicit invocation schema generation: `python invocationSchemaHandler.py -i toolDescriptor.json -d tool.exampleInputs.json`
+    * Validate input data (i.e. specific tool parameter arguments): `python invocationSchemaHandler.py -i toolDescriptor.json -d tool.exampleInputs.json`
   * Check the help page (`-h` option) for more options and documentation.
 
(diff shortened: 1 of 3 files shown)


tools/README.md
```diff
@@ -19,5 +19,12 @@ Note: there is a Docker image ([boutiques/boutiques](https://hub.docker.com/r/bo
     * Print a command line based on parameters from an input file: `docker run --rm -v $PWD:/work -w /work boutiques/boutiques localExec.py -i input_param_file.csv ./tool.json`
   * Check the help page (`-h` option) for more options and documentation.
 
-3. `pegasus-boutiques`: a python API to generate [Pegasus](https://pegasus.isi.edu) workflows from Boutiques JSON descriptors.
+3. `invocationSchemaHandler.py`: a Python script to generate and test invocation schemas (i.e. JSON schemas for input values) associated to a Boutiques application
+     descriptor and validating input data examples with respect to these invocation schemas.
+  * Usage examples:
+    * Generates an invocation schema: `python invocationSchemaHandler.py -i toolDescriptor.json`
+    * Validate input data (i.e. tool parameter arguments) with implicit invocation schema generation: `python invocationSchemaHandler.py -i toolDescriptor.json -d tool.exampleInputs.json`
```
glatard: "Implicit invocation schema generation" is unclear. I'd say "validate input data written in the invocation schema".

ttaa9: That's also a little confusing to me because the invocation schema in that case will not be displayed or written, it be generated and used to check the specific input values in `tool.exampleInputs.json`.

How about just "validate input data"?

ttaa9:
I've added the package in docker and changed the readme wording

tools/README.md
glatard: yep

PR #75, opened 2016-10-24
uses-absolute-path for output files

ttaa9:
Adds the uses-absolute-path property to the schema, readme, and localexec.
Also, fixes some bugs in the localexec.


glatard:
Great!

PR #78, opened 2016-10-25
Updates for Boutiques container

glatard:


ttaa9:
Oops, forgot to add `run`; thanks for that.

PR #84, opened 2016-10-27
Support for configuration files

glatard:
- Configuration files are output files (yes, _output_ and not _input_ because they are produced by the execution and cannot be specified by the user) that have a `file-template`.
- `file-template` is the template of the configuration file. It may contain input and output keys that will be substituted at runtime as if they were on the command line (except that flags are not used and when they have no value the whole line of the file is suppressed).
- `file-template` are expressed as arrays of strings, each string representing a line of the file. This is because JSON doesn't support multi-line strings. 
- `command-line-key` has been renamed into `value-key` accordingly.

P.S.: sorry, this PR contains the commits in #83 (I know, it's bad). Please review #83 before this one.


ttaa9:
This does seem like the best way to do config files. 

One thought I initially have is whether config files should be separate from output (and input) files. Semantically speaking, the config file is not an output file of the tool or application (it's an input in that sense), but rather one of the platform. Of course, practically speaking it's not that important. But down the road if we want many properties specific to output files vs config files, though, it might make a difference.

schema/README.md
```diff
@@ -20,10 +20,10 @@ The format of command line keys is not specified. However, it is recommended to
   * **`name`:** a human-readable input name. Example: "Data File".
   * **`type`:** input type. "File", "String", "Flag", "Enum", or "Number". Type "File" also includes directories.
   * `description`: input description.
-  * `command-line-key`: a string, contained in `command-line`, substituted by the input value and/or flag at runtime.
+  * `value-key`: a string, contained in `command-line`, substituted by the input value and/or flag at runtime.
```
ttaa9: Maybe add "...and/or in `file-template`"

schema/README.md
```diff
@@ -40,14 +40,15 @@ The format of command line keys is not specified. However, it is recommended to
 * **`output-files`**: an array of objects that represent output files with the following properties:
   * **`id`:** a short, unique, informative identifier containing only alphanumeric characters and underscores. Typically used to generate variable names. Example: "data_file".
   * **`name`**: output name.
-  * **`path-template`**: a string that describes the output file path relatively to the execution directory. May contain input `command-line-keys`. Example: ```results/[INPUT1]_brain.mnc```.
+  * **`path-template`**: a string that describes the output file path relatively to the execution directory. May contain input `value-keys`. Example: ```results/[INPUT1]_brain.mnc```.
   * `description`: output description.
   * `path-template-stripped-extensions`: list of file extensions that will be stripped from the input values before being substituted in the path template. Example: ```[".nii",".nii.gz"]```.
-  * `command-line-key`: a string, contained in `command-line`, substituted by the output value/flag at runtime.
+  * `value-key`: a string, contained in `command-line`, substituted by the output value/flag at runtime.
```
ttaa9: Again note it can be in `file-template`?

schema/descriptor.schema.json
```diff
@@ -360,7 +360,7 @@
                         }
                     },
                     "exclusive-minimum": ["minimum"],
-                    "exclusive-maximum": ["maximum"]
+                    "exclusive-maximum": ["maximum"]
```
ttaa9: Trailing space?

schema/descriptor.schema.json
```diff
@@ -421,21 +421,43 @@
...
                         "type": "string"
                     },
                     "uses-absolute-path": {
                         "id": "http://github.com/boutiques/boutiques-schema/output-files/uses-absolute-path",
                         "description": "Specifies that this output filepath will be given as an absolute path.",
                         "type": "boolean"
+                    },
+                    "file-template": {
+                        "id": "http://github.com/boutiques/boutiques-schema/input/file-template",
+                        "description": "An array of strings that may contain command line keys. Each item will be a line in the configuration file.",
```
ttaa9: Do you want to call `command line keys` here `value keys` instead?

tools/localExec.py
```diff
@@ -31,10 +31,12 @@ def __init__(self,desc,options={}):
     ## Helpers Functions ##
     # The set of input parameters from the json descriptor
     self.inputs     = self.desc_dict['inputs'] # Structure: [ {id:...},...,{id:...} ]
+    # The set of output parameters from the json descriptor
+    self.outputs     = self.desc_dict['output-files'] # Structure: [ {id:...},...,{id:...} ]
     # The set of parameter groups, according to the json descriptor
     self.groups     = self.desc_dict['groups'] if 'groups' in self.desc_dict.keys() else []
     # Retrieves the parameter corresponding to the given id
-    self.byId       = lambda n: [ v for v in self.inputs if v['id']==n][0]
+    self.byId       = lambda n: [ v for v in self.inputs+self.outputs if v['id']==n][0]
```
ttaa9: This one worries me a bit, since previously I assumed I could only get inputs from it. But, looking through the script, I think it is ok.

examples/configuration-file/config.json
```diff
@@ -0,0 +1,66 @@
...
+	"id": "optional_input",
+	"name": "Optional input",
+	"type": "String",
+	"value-key": "[OPTION]",
+	"optional": true
+    }],
+    "output-files": [{
+	"id": "output_file",
+	"name": "Output file",
+        "value-key": "[OUTPUT-FILE]",
```
glatard: Doesn't show in my editor or in https://github.com/glatard/boutiques/blob/02928966fc5e215dfd897a669c5608cad6b38438/examples/configuration-file/config.json

examples/configuration-file/config.json
```diff
@@ -0,0 +1,66 @@
...
+	"path-template": "[CONFIG_FILE]-[FILE-INPUT].o",
+	"path-template-stripped-extensions": [".txt"]
+    },{
+	"id": "config_file",
+	"name": "Configuration file",
+	"type": "Configuration File",
+	"value-key": "[CONFIG_FILE]",
+        "path-template": "config.txt",
```
glatard: Same comment.

schema/README.md
glatard: fixed.

schema/descriptor.schema.json
glatard: fixed

schema/descriptor.schema.json
glatard: yes, good catch.

glatard:
@ttaa9, thanks for the review. I just pushed some fixes following your comments. I agree with you that it's a bit weird to consider that configuration files are output files. However, if we separate them in a distinct object (an array of configuration files), all the properties in output-files except `list` would have to be duplicated. Do we want to go for this?

Issue #79, opened 2016-10-26
Set additionalProperties to false wherever possible


glatard:
Add a "customProperties" array in the main object to allow for custom properties. In the documentation, recommend that custom properties start with "[platformName]:".

Issue #72, opened 2016-10-19
Corral custom properties under a common key 


glatard:
Make `custom` an optional key that must be an object. Then set `additionalProperties` to false. As per @kofalt's suggestion in #66.

Issue #69, opened 2016-10-18
Add invocation schema to JSON schema.


glatard:
As an optional property.

PR #86, opened 2016-11-03
Schema updates

glatard:
Schema updates to address #69, #72 and #79, with corresponding examples and documentation.

### DIFF
--- a/schema/descriptor.schema.json
+++ b/schema/descriptor.schema.json
@@ -62,9 +62,12 @@
                             "description": "Index where the image is available.",
                             "default": "http://index.docker.io",
                             "type": "string"
-                        }
+                        },           
+                        "working-directory": {},
+                        "container-hash": {}
                     },
                     "required": ["type", "image"],
+                    "additionalProperties": false,
                     "description": "Object representing a docker container for the tool."
                 }, {
                     "properties": {
@@ -74,9 +77,12 @@
                             "minLength": 1,
                             "description": "URL where the image is available.",
                             "type": "string"
-                        }
+                        },
+                       "working-directory": {},
+                       "container-hash": {}
                     },
                     "required": ["type", "url"],
+                     "additionalProperties": false,
                     "description": "Object representing a rootfs or singularity container for the tool."
                 }]
             }]
@@ -85,7 +91,7 @@
             "id": "http://github.com/boutiques/boutiques-schema/schema-version",
             "type": "string",
             "description": "Version of the schema used.",
-            "enum": ["0.2", "0.3"]
+            "enum": ["0.4"]
         },
         "walltime-estimate": {
             "id": "http://github.com/boutiques/boutiques-schema/walltime-estimate",
@@ -122,7 +128,8 @@
                 "required": [
                     "name",
                     "value"
-                ]
+                ],
+                "additionalProperties": false
             }
         },
         "groups": {
@@ -176,7 +183,8 @@
                     "name",
                     "id",
                     "members"
-                ]
+                ],
+                "additionalProperties": false
             }
         },
         "inputs": {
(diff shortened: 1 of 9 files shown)


glatard:
@kofalt may have an opinion on this too.

schema/descriptor.schema.json
```diff
@@ -85,7 +91,7 @@
             "id": "http://github.com/boutiques/boutiques-schema/schema-version",
             "type": "string",
             "description": "Version of the schema used.",
-            "enum": ["0.2", "0.3"]
+            "enum": ["0.4"]
```
kofalt: So this raises an interesting question - if a system wants backwards compatibility with multiple boutique versions, would the approach have to be first, take each released json schema to see which (if any) match, and go from there?

My intuition says "yes"; curious what you think :smile:

kofalt:
Changes look positive, I like it. 

I have a comment on #69 that I should have written on that ticket & is my fault for not writing earlier. Seeing how the invocation schema is a 100% generated field, it should be A) always possible to generate from a boutique doc, B) possible to have an error in what the author wrote!

For these reasons I'm tempted to think of the invocation schema as a layer above the alg specification. This is our current approach in the Exchange for both [boutiques](https://github.com/flywheel-io/exchange/blob/master/manifests/alpine-sha384-159fa0a4649f1eb5c9603fdacbcf4e2ff1676dce8b163b7925a0b83613b8f49e38b4be708df063a15b28df3db3988c71.json) and [gears](https://github.com/flywheel-io/exchange/blob/master/manifests/scitran/scitran-dcm2niix-sha384-e61739c42fcbd5370f15c380860efd5b65a5961b81fb6b110305dc5195c5c09c668770cd0c897731ebe5ef228cefc1e2.json); we generate the invocation schema when the Exchange receives your document and stores it - separately - for convenience and archival purposes.

This means that the only stored invocation schema is generated by a submission process that will do it the same time every way, and removes the question of "what do I do when a boutique has an invalid or non-standard invocation schema".

TL;DR I would support **not** having an invocation schema on the boutique itself and it's the system's responsibility to either generate it on-the-fly or at the point of archival, to remove possible errors & confusion. Thoughts?

schema/descriptor.schema.json
kofalt: Sounds good to me :+1:

glatard:
I understand your point but, although we haven't played a lot with invocation schemas so far, I see a few issues:
* Generating the invocation schema requires a specific tool, currently the invocationSchemaHandler.py. It is much easier for a platform to validate inputs using straight JSON schema validation (which can be done in any language) than to have to integrate a Python script somewhere in the process. Ideally, invocation schemas should be generated asap, ideally every time there is an edit in the manifest (that's why it's so cool to have a repo that can trigger such operations). If platforms have to do it, it will only add complexity.
* Keeping the schema and manifest separated will create issues when both documents are inconsistent (what if you're testing 2 versions of the same manifest with incompatible schemas, etc).

That's why I think including the invocation schema in the manifest should be the norm. However, if for some reasons a platform wants to ignore this property and re-generate the invocation schemas to store them separately (or extract the schemas and store them in a different document), it is always possible.

What do you think?

kofalt:
My main concern is that an author adds some option (or something) and forgets to update his invocation - when I'm processing that boutique, do I honor their (invalid) schema? Generate my own? Why even have that possibility? It's the difference between the system being invalid: go debug it, or the author being invalid: send an email :/

Not needing python to run a boutique is a good feature though, it's one of the reasons the Exchange generates the invocation & stores it. If this lands, maybe we should update the Exchange will reject boutiques that are submitted with a schema that differs from that spec version?

I would be fine going ahead with this as long as it's clear that the invocation should always match what we expect to generate, and that consumers have the option of using their own variant of the invocation, which is generally what I would do :smile:

glatard:
I agree that inconsistencies between the manifest and the invocation schema could still be a problem. If I received a manifest with an invocation schema from a developer (human), I would definitely double-check it.

Given that the handling of invocation schemas is still a bit nascent, I would suggest that we keep this property for now, and that

kofalt:
> for now, and that 

I think your comment cut off :) But yeah, I'm cool with that.

Issue #59, opened 2016-09-13
Enum should not be an input type


johnflavin:
Rather than having a special input type that allows me to restrict the values of the input to a set, I should be able to restrict the values of _any_ input type to a set. `Enum` should not be an input type, there should be an optional property that any input type can define which will restrict the valid values to a set.

I propose to rename the input property `enum-value-choices` to `value-choices`, and remove the dependency that `enum-value-choices` can only be set for `type=Enum`. (I guess we would also have to remove `items: {type: string}` from `value-choices`? Not sure about that.)

So under this proposed change, an input could be something like 

``` json
{
    "id": "foo",
    "type": "String",
    "value-choices": ["a", "b", "c"]
}
```

Similarly,

``` json
{
    "id": "bar",
    "type": "Number",
    "value-choices": [0, 1, 2]
}
```

should be possible.

glatard:
Yes, I tend to agree with that. @ttaa9?

ttaa9:
Seems reasonable to me. There are cases where I indeed had to put ["1", "2", "3"], although I didn't consider that an issue at the time.

PR #90, opened 2016-11-11
Fixes for #59 and #79

glatard:
@johnflavin, you may want to have a look too (for #59).

### DIFF
--- a/schema/descriptor.schema.json
+++ b/schema/descriptor.schema.json
@@ -431,6 +431,11 @@
                         "description": "Option flag of the output, involved in the value-key substitution. Examples: -o, --output",
                         "type": "string"
                     },
+		    "command-line-flag-separator": {
+                        "id": "http://github.com/boutiques/boutiques-schema/output/command-line-flag-separator",
+                        "description": "Separator used between flags and their arguments. Defaults to a single space.",
+                        "type": "string"
+                    },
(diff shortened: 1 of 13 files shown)


schema/descriptor.schema.json
```diff
@@ -324,16 +327,16 @@
...
                 }],
                 "dependencies": {
                     "command-line-flag-separator": ["command-line-flag"],
                     "min-list-entries": ["list"],
                     "max-list-entries": ["list"],
-                    "enum-value-choices": {
+                    "value-choices": {
                         "properties": {
-                            "type": {"enum": ["Enum"]}
+                            "type": {"enum": ["String", "Number"]}
```
ttaa9: You don't want to include `File`?

ttaa9:
Looks good to me. The only issue is that `validator.rb` and `invocationSchemaHandler.py` are not up-to-date with respect to these enum changes.

schema/descriptor.schema.json
glatard: Not sure how we could specify a list of values for file... Should it be a list of URLs? Anyway, let's leave it aside for now.

glatard:
@ttaa9 I just updated the two scripts. Please pay particular attention to invocationSchemaHandler.py as I'm a bit unsure of what I did (though it seems to be working with the example).

Issue #80, opened 2016-10-26
Move as many checks as possible from validator.rb to JSON schema


glatard:
For instance, the following checks are easy to add to the schema:
* Flag inputs have command line flags
* Flag inputs must not be required
* Flag inputs cannot be lists
* Enum types always have at least 1 choice (see #70 though) 
* Non-list inputs cannot have the min/max list entries property

Other ones (e.g. check that an input cannot require and disable the same input) look trickier but it still might be possible to include them in the schema. 

The goal is to have as little information as possible in the validator so that platforms could rely on simple JSON validation using whatever language, instead of having to add some code.

gkiar:
@glatard is this obsolete now, or no?

glatard:
It is not. We still do many things in Python that we could probably add to the schema.

PR #104, opened 2017-07-25
added validator in python

gkiar:
## New Features
- Python package structure
- Pip-installable for dev with `pip install . -e` from within `/tools/python/` directory; not yet on pypi
- Validator now exists in Python (test of operation below)

## Reorganization
- Moved schema within Python package, added symbolic link to this file from previous location
- Organized tools folder based on language
- Added a failure and success example to the schema folder in the Python package, enabling Python testing later on

### Success Mode
There are no differences in output between the implementations in the success mode.

**Ruby**:
```
$ ruby ruby/validator.rb python/boutiques/schema/descriptor.schema.json python/boutiques/schema/example_good.json
["OK"]
```
**Python**:
```
$ python boutiques/validator.py boutiques/schema/example_good.json
OK
```

### Failure Mode
There are two minor differences between the implementations. They are as follows:
- The Python implementation provides labels for the errors, enabling easier interpretation and correction
- The Python implementation detected that the key "ey2" is contained within "key2," whereas the Ruby implementation did not

**Ruby**:
```
$ ruby ruby/validator.rb python/boutiques/schema/descriptor.schema.json python/boutiques/schema/example_bad.json
["[ey2] not in cmd line or file template",
 "Non-unique id key2",
 "Output files out1 and out2 have the same path-template",
 "Required id key8 for key4 was not found",
 "Required param key4 cannot require other inputs",
 "key5 cannot have an default value outside its choices",
 "Required param key5 cannot require other inputs",
 "key7 cannot have greater min (1) than max (-1)",
 "Required id nonkey1 for key7 was not found",
 "Disabled id nonkey1 for key7 was not found",
 "Id key7 requires and disables nonkey1",
 "flag1 must have a command-line flag",
 "flag1 should not be required",
 "key6 cannot appear twice (in groups group1 & group1)",
 "key7 cannot appear twice (in groups group1 & group2)",
 "key5 in mutex group group1 cannot be required",
 "key5 in mutex group group1 cannot require fellow member key6",
 "key5 in one-is-required group group1 should not be required",
 "Member id key8 from group group2 is not present in the inputs"]
```
**Python**:
```
$ python boutiques/validator.py boutiques/schema/example_bad.json
$ python boutiques/validator.py boutiques/schema/example_bad.json
   KeyError: "ey2" not in command-line or file template
   KeyError: "key2" contains "ey2"
    IdError: "key2" is non-unique
OutputError: "out1" and "out2" have the same path-template
 InputError: "key4" required id "key8" not found
 InputError: "key4" cannot require or disable other inputs
 InputError: "key5" cannot have default value outside its choices
 InputError: "key5" cannot require or disable other inputs
 InputError: "key7" cannot have greater min (1) than max (-1)
 InputError: "key7" required id "nonkey1" not found
 InputError: "key7" disabled id "nonkey1" not found
 InputError: "key7" requires and disables "nonkey1"
 InputError: "flag1" must have a command-line flag
 InputError: "flag1" should not be required
 GroupError: "group1" member "key6" appears twice
 GroupError: "group1" is mutually-exclusive and cannot have required members, such as "key5"
 GroupError: "group1" is mutually-exclusive and cannot have members require one another, such as "key5" and "key6"
 GroupError: "group1" is a one-is-required group and contains a required member, "key5"
 GroupError: "group2" member "key8" does not exist
 GroupError: "group2" and "group1" both contain member "key7"
```


tools/python/boutiques/invocationSchemaHandler.py
```diff
@@ -3,7 +3,12 @@
 # A script for handling invocation schema generation and validation for Boutiques application descriptors
 # Requires jsonschema 2.5
 
-import json, jsonschema as jsa, argparse, os, sys
+import json
+import jsonschema
```
natacha-beck: Take a look here for your issue (https://github.com/boutiques/boutiques/issues/105). You don not have anymore `jsa` defined, you should use `jsonschema` instead.

schema/descriptor.schema.json
```diff
@@ -1,501 +0,0 @@
-{
```
glatard: If it doesn't harm the Python package, I would make the symbolic link in the other direction (file in `schema`, link in `tools/python`) just in case other languages need to refer to it.

tools/python/boutiques/__init__.py
```diff
@@ -0,0 +1,10 @@
+from __future__ import absolute_import
+
+from .localExec import LocalExecutor
+from .invocationSchemaHandler import generateInvocationSchema, writeSchema
+from .validator import validate_json
+
+__all__ = ['localExec', 'invocationSchemaHandler', 'validator']
+
+version="0.0.1"
```
glatard: Shouldn't this be the next tag in the repository? (I guess it would be 0.4.1, the schema being still 0.4).

tools/LICENSE
```diff
@@ -1,340 +0,0 @@
-GNU GENERAL PUBLIC LICENSE
```
glatard: yeah, good catch!

tools/python/boutiques/localExec.py
```diff
@@ -34,46 +34,45 @@ def __init__(self,desc,options={}):
     # The set of output parameters from the json descriptor
     self.outputs     = self.desc_dict['output-files'] # Structure: [ {id:...},...,{id:...} ]
     # The set of parameter groups, according to the json descriptor
-    self.groups     = self.desc_dict['groups'] if 'groups' in self.desc_dict.keys() else []
+    self.groups     = self.desc_dict['groups'] if 'groups' in list(self.desc_dict.keys()) else []
```
glatard: not sure why this list and the two other ones below has to be created but fair enough.

tools/python/boutiques/localExec.py
```diff
@@ -96,18 +95,18 @@ def execute(self):
...
       dcmd = 'docker run --rm' + envString + ' -v ${PWD}:' + launchDir + ' -w ' + launchDir + ' ' + str(dockerImage) + ' ./' + dsname
-      print('Executing in Docker via: ' + dcmd)
+      print(('Executing in Docker via: ' + dcmd))
       exit_code = self._localExecute( dcmd )
     # Otherwise, just run command locally
     else:
       exit_code = self._localExecute( command )
     # Report exit status
-    print('---/* End program output */---\nCompleted execution (exit code: ' + str(exit_code) + ')')
+    print(('---/* End program output */---\nCompleted execution (exit code: ' + str(exit_code) + ')'))
```
glatard: for my own culture, why are such double parentheses useful?

tools/python/boutiques/schema/descriptor.schema.json
```diff
@@ -0,0 +1,501 @@
+{
```
glatard: how about using `git mv` to preserve the history of the file? (ignore this comment if you did it, github seems to show it as a new file...)

tools/python/boutiques/schema/example_bad.json
```diff
@@ -0,0 +1,138 @@
+{
+  "name": "demo-tool",
+  "tool-version": "v1",
+  "description": "a tool that does things",
+  "command-line": "ndmg_pipeline [key1] [key2] [key3] [key4] [key5] [key6] [key7] [flag1]",
+  "container-image": {
+    "type": "docker",
+    "image": "user/image:v1",
+    "index": "http://index.docker.io"
```
glatard: you could add a `url` property here (not allowed with type `docker`.)

tools/python/boutiques/schema/example_good.json
```diff
@@ -0,0 +1,236 @@
...
+      "optional": true,
+      "value-choices": [
+        "/ndmg_atlases/labels/HarvardOxford.nii.gz"
+      ]
+    },
+    {
+      "id": "cpac200",
+      "name": "CPAC200 parcellation",
+      "type": "String",
+      "value-key": "[CPAC200]",
```
glatard: you could name one of these `command-line-key` as this was the term up until config files were introduced and some descriptors may still use it. (see `invalid-tool.json` in `examples/invalid/typos`)

tools/python/boutiques/schema/example_good.json
```diff
@@ -0,0 +1,236 @@
+{
+  "name": "ndmg",
+  "tool-version": "v0.0.41-1",
+  "description": "dwi connectome estimation pipeline",
+  "command-line": "ndmg_pipeline [DTI] [BVAL] [BVEC] [MPRAGE] [ATLAS] [MASK] [OUTDIR] [DESIKAN] [JHU] [TALAIRACH] [AAL] [HARVARDOXFORD] [CPAC200] [CLEAN]",
+  "container-image": {
```
glatard: you could spell this `containerImage` to make sure that this is not silently ignored (see `invalid-tool.json` in `examples/invalid/typos`).

tools/python/boutiques/validator.py
```diff
@@ -0,0 +1,221 @@
...
+        # Verify flag-type inputs (have flags, not required, cannot be lists)
+        if inp["type"] == "Flag":
+            msg_template = " InputError: \"{}\" must have a command-line flag"
+            errors += [msg_template.format(inp["id"])] if "command-line-flag" not in inp.keys() else []
+
+            msg_template = " InputError: \"{}\" should not be required"
+            errors += [msg_template.format(inp["id"])] if inp["optional"] is False else []
+
+            # This one is redundant as basic JSON validation catches it
+            msg_template = " InputError: \"{}\" cannot be a list"
```
glatard: This should now be caught by the schema, could you verify and remove the code if confirmed? Catching that flags aren't required and have a command line flag should be feasible in the schema, by adding to the object in line 326 of the current schema (but let's leave it for another PR). See also #80.

tools/python/boutiques/validator.py
```diff
@@ -0,0 +1,221 @@
...
+            errors += [msg_template.format(inp["id"], minn, maxx)] if minn > maxx else []
+
+            msg_template = " InputError: \"{}\" cannot have negative min entries ({})"
+            errors += [msg_template.format(inp["id"], minn)] if minn < 0 else []
+
+            msg_template = " InputError: \"{}\" cannot have non-positive max entries ({})"
+            errors += [msg_template.format(inp["id"], maxx)] if maxx <= 0 else []
+
+        # Verify non list-type inputs don"t have list properties
+        # This one is redundant as basic JSON validation catches it
```
glatard: Then let's remove it!

tools/python/setup.py
```diff
@@ -0,0 +1,17 @@
+from setuptools import setup
+
+setup(name='boutiques',
+      version='0.1',
```
glatard: Same comment as before on version.

tools/python/setup.py
```diff
@@ -0,0 +1,17 @@
...
+      description='Schema for describing commandlines',
```
glatard: `bash` command-line tools.

tools/python/setup.py
```diff
@@ -0,0 +1,17 @@
...
+      url='http://github.com/boutiques/boutiques',
+      author='Tristan Glatard',
+      author_email='tristan.glatard@gmail.com',
+      license='GPL',
```
glatard: It's now MIT.

tools/python/boutiques/localExec.py
gkiar: These were added in the 2to3 conversion, so that it reliably works consistently with both. Getting keys in Python3 gives you a `dict_keys` object, not a list.

tools/python/boutiques/__init__.py
gkiar: Agreed! Good catch

schema/descriptor.schema.json
gkiar: I prefer that as well, but it does seem to break the python package - it can only contain (to my knowledge) files that are downstream of the main directory containing `setup.py`. One option would be removing the validator in Ruby, as you said, and having the Python directory sit at the base of the repo. Then we could have the schema separate from the package, but still include it as it'd be downstream. Thoughts?

tools/python/boutiques/localExec.py
gkiar: That was a mistake in the 2to3 conversion that I failed to catch - I'll remove the second set :) Thanks!

tools/python/boutiques/schema/descriptor.schema.json
gkiar: Ya, I think that might be because there is a symbolic link with that original filename/path, so the history got confused. I'll try to fix it

tools/python/boutiques/validator.py
gkiar: This one is caught by the schema, I just wrote this test in to exhaustively write-out the checks that were in the Ruby validator. One other never gets triggered as the schema catches it up stream, as well. I'll remove the code :)

tools/python/boutiques/schema/descriptor.schema.json
gkiar: I reverted the files and redid this move, ensuring I did `git mv` - it appears like no dice locally, but maybe when it's merged the history changes will have taken effect.

tools/python/boutiques/schema/example_good.json
gkiar: Tested this - the typo correctly raises an error in the schema. I want `example_good.json` to be an example that passes, so won't propagate this change to the online version.

tools/python/boutiques/schema/example_bad.json
gkiar: Tested - successfully caught by JSON validation. I don't want to push this change (ones directly flagged by the schema) to Github in this file as it is designed to pass schema validation but fail extended full Boutiques validation.

tools/python/boutiques/schema/example_good.json
gkiar: Do you want the schema to be backwards compatible and support this key, or not? If so, I added it as well to the schema and can push that, but if you want it to be caught as an error, without any changes it was :)

gkiar:
Updated PR to aim at the `develop` branch, made all suggested code adjustments, and tested all suggested tests. 😄 

I explained in comments above why I didn't propagate all the tests to files online - generally, I wanted to have the example in the package test the validator, not the schema (i.e. changes that the schema caught were not tested here, as in the event of general schema failure, Boutiques-specific validation does not continue).

Only outstanding question I can think of @glatard is whether or not you want Boutiques to be backwards compatible with the `command-line-key` field as well as `value-key`? I have made these changes to the schema and validator locally, but have not propagated yet as I wanted to know your answer before doing so.

PR #109, opened 2017-08-28
Validator can now parse descriptors with input that don't have any 'o…

glatard:
…ptional' property (defaults to false).


gkiar:
@glatard do we want to try and conform to a style guide? For instance, PEP8 (which is what I try to mostly follow) requires spaces before and after `=`

Issue #106, opened 2017-08-09
In the schema, support container images stored in SingularityHub


glatard:
Docker containers could be defined from URL on DockerHub instead of image name, to make it consistent with Singularity.

gkiar:
Addressed in https://github.com/boutiques/boutiques/pull/113; note that this breaks backwards compatibility and should be addressed in the release.

PR #113, opened 2017-08-28
Schema & localExec (i.e. bosh) updates

gkiar:
- Added entrypoint flag to schema
- Moved `singularity` to be handled like `docker` rather than `rootfs`
- Generalized `localExec.py` for singularity and docker containers

### DIFF
--- a/tools/python/boutiques/schema/descriptor.schema.json
+++ b/tools/python/boutiques/schema/descriptor.schema.json
@@ -70,7 +70,7 @@
                         "working-directory": {},
                         "container-hash": {}
                     },
-                    "required": ["type", "image", "entrypoint"],
+                    "required": ["type", "image"],
                     "additionalProperties": false,
                     "description": "Object representing a docker or singularity container for the tool."
                 }, {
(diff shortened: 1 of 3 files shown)


tools/python/boutiques/schema/descriptor.schema.json
```diff
@@ -49,29 +49,33 @@
...
-                            "description": "Index where the image is available.",
-                            "default": "http://index.docker.io",
+                            "description": "Optional index where the image is available, if not the standard location. Example: docker.io",
                             "type": "string"
-                        },           
+                        },
                         "working-directory": {},
                         "container-hash": {}
                     },
-                    "required": ["type", "image"],
```
glatard: how about making entrypoint optional? It would default to false, which would make the minimal descriptor less verbose. And not break backward compatibility.

tools/python/boutiques/schema/descriptor.schema.json
```diff
@@ -49,29 +49,33 @@
             },{
                 "oneOf": [{
                     "properties": {
-                        "type": { "enum": ["docker"] },
+                        "type": { "enum": ["docker", "singularity"] },
```
glatard: This breaks backward compatibility, not a problem but we should mention it in the release notes.

gkiar: Good point - I added a comment in https://github.com/boutiques/boutiques/issues/106

PR #145, opened 2017-09-19
Added support for directory mounting in containers (Fixes #142)

### LINKED ISSUE #142: Clean Boutiques examples
Update them to version 0.5
Add them to tests

glatard:
Usage example: ```./localExec.py --mounts /tmp:/tmp /a:/b -i ./invocation.json ./example1.json -e```


tools/python/boutiques/localExec.py
```diff
@@ -110,10 +109,17 @@ def execute(self):
       launchDir = '${PWD}' if (self.launchDir is None) else self.launchDir
       # Run it in docker
       if conType == 'docker':
-        dcmd = 'docker run --entrypoint=/bin/bash --rm' + envString + ' -v ${PWD}:' + launchDir + ' -w ' + launchDir + ' ' + str(conImage) + ' ./' + dsname
+        # export mounts to docker string
+        docker_mounts = ""
+        for mount_string in mount_strings:
+          docker_mounts += " -v "+mount_string
```
gkiar: From this it is unclear to me exactly how we expect the strings to be passed. Intuitively I almost expect a list like the following:
`[/local/path1:/data1, /local/path2:/data2, ...]`

If that's the case, then we could just add `'${PWD}:'+launchDir` to this list, and join them in the string with a method like: `" -v ".join([dm for dm in docker_mounts])`

What do you think? The benefit in my head is that it's just internally consistent for mounts.

tools/python/boutiques/localExec.py
```diff
@@ -110,10 +109,17 @@ def execute(self):
...
+        dcmd = 'docker run --entrypoint=/bin/bash --rm' + envString + docker_mounts + ' -v ${PWD}:' + launchDir + ' -w ' + launchDir + ' ' + str(conImage) + ' ./' + dsname
       elif conType == 'singularity':
+        singularity_mounts = ""
+        for mount_string in mounts:
+          singularity_mounts += " -B "+mount_string
         #TODO: Test singularity runtime on cluster
-        dcmd = 'singularity exec --rm' + envString + ' -v ${PWD}:' + launchDir + ' -w ' + launchDir + ' ' + str(conImage) + ' ./' + dsname
+        dcmd = 'singularity exec --rm' + envString + singularity_mounts + ' -B ${PWD}:' + launchDir + ' -w ' + launchDir + ' ' + str(conImage) + ' ./' + dsname
```
gkiar: same comment+question as above

tools/python/boutiques/localExec.py
```diff
@@ -594,8 +600,9 @@ def main():
   parser.add_argument('-r', '--random', action = 'store_true', help = 'Generate a random set of input parameters to check.')
   parser.add_argument('-n', '--num', type = int, help = 'Number of random parameter sets to examine.')
   parser.add_argument('-s', '--string', help = "Take as input a semicolon-separated string of comma-separated tuples on the command line.")
+  parser.add_argument('-v', '--mounts', type=str, nargs="*", help = "Directories to be mounted in the container in addition to $PWD. Will be passed to Docker with -v or to Singularity with -B.")
```
gkiar: What format are we currently expecting these inputs in? I think our help-text should specify.

tools/python/boutiques/localExec.py
glatard: yes, sure, of course!

glatard:
@gkiar: this is ready for another review!

glatard:
@gkiar, wait a sec, I found a bug while writing tests :)

glatard:
ok, good to go!

tools/python/boutiques/localExec.py
```diff
@@ -109,11 +108,16 @@ def execute(self):
       # Change launch (working) directory if desired
       launchDir = '${PWD}' if (self.launchDir is None) else self.launchDir
       # Run it in docker
+      mount_strings = [] if not mount_strings else mount_strings
+      mount_strings.append('${PWD}:' + launchDir)
       if conType == 'docker':
-        dcmd = 'docker run --entrypoint=/bin/bash --rm' + envString + ' -v ${PWD}:' + launchDir + ' -w ' + launchDir + ' ' + str(conImage) + ' ./' + dsname
+        # export mounts to docker string
+        docker_mounts = " -v ".join(m for m in mount_strings)
+        dcmd = 'docker run --entrypoint=/bin/bash --rm' + envString + ' -v '+docker_mounts+ ' -w ' + launchDir + ' ' + str(conImage) + ' ./' + dsname
```
gkiar: Can you confirm if we need the `-v` on this line since we add it to the `docker_mounts` variable above? I never remember if the join prepends the separator as well. (Would test but on my phone, will test when back at a computer if you haven't replied yet)

PR #147, opened 2017-09-21
bosh testing

### LINKED ISSUE #88: localExecutor crashes when one of the inputs has no optional property
In JSON, the default value of boolean properties is false.

glatard:
* Moved and updated `example1` to use it in tests.
   * Partially addresses #142 
   * Requires that Travis has Docker installed. I haven't configured it.
   * Question: the tests currently pull `boutiques/example1` from DockerHub, through `bosh`. Do we want to build `boutiques/example1` during the tests and if yes, how? 
* Added tests for `bosh`.
* Fixed #88.
* Fixed validator crash when descriptor has a config file. 
* Includes #145, sorry... let's merge #145 first.


gkiar:
To summarize/document chatter in person:
 - this is awesome
 - we do want to build the container locally
 - we need to setup docker in travis
 - the above comment in singularity is a bug
 - @gkiar addressed all of the above in a PR back to the @glatard fork, that once merged remotely, will update this PR (https://github.com/glatard/boutiques/pull/1)

PR #153, opened 2017-09-27
LocalExec updates

### LINKED ISSUE #141: bosh: support inputs complying to invocation schema
* Remove other input format types.
* Tool could generate empty invocation to help get started.

glatard:
* Fixes #141 
*  Removes ignore containers option
* Makes destroy temp scripts default


glatard:
* we can clean up the interface too. I just removed two options here that I thought weren't useful. Simplification is always good! Maybe in a next PR though.
* yes, I removed the invocation schema from good.json so that bosh-invocation regenerates it. If a schema is already in the descriptor, bosh-invocation would just grab it and not regenerate it, which reduces test coverage!

gkiar:
Awesome, on both counts. Testing this locally now before I approve :)

Issue #133, opened 2017-09-08
schema: add more information about tools


glatard:

glatard:
Add `tool_author` (could be URL to contributors), `tool_url` and `tool_doi` at top level. Once done, update `bosh-publish` to use these properties.

glatard:
Schema is here: https://github.com/boutiques/boutiques/blob/master/tools/python/boutiques/schema/descriptor.schema.json

PR #159, opened 2017-10-02
Update and Centralization of API

gkiar:
## Proposed Command-line API
I suggest we have, as discussed, a two-level API for the `bosh` commands. This means that `bosh -h` will simply provide help as to which options are available (i.e. `['validate', 'exec', 'import', 'publish', 'invocation']`), and when one of them is provided the `-h` option (or misspecification of other options required by the module) will be handled by a separate argument parser, so that modes do not contaminate one another.

I propose we do a third layer for `localExec` (with the split happening at `launch` vs `simulate` modes), as well. This is not necessary for `validate`, `import`, `invocation`, or `publish`.

Things that will be consistent across all tools:
- Names of common parameters
- Order of required parameters
- Short-hands for flags, where appropriate (these will also seek to be consistent with similar tools in Unix)
- The command-line and proposed operation from within Python will be identical

### Tool CLIs 
| Tool | Usage | 
|:-----|:-----------|
| Validator | `bosh validate descriptor.json {-b}` |
| LocalExec (launch) | `bosh exec launch descriptor.json inputs.json` |
| LocalExec (simulate) | `bosh exec simulate descriptor.json {-i inputs.json, -r, -n #}` |
| InvocationSchema | `bosh invocation descriptor.json -i inputs.json` |
| Publish | `bosh publish {all-the-many-args}` |
| Import | `bosh import {0.4, bids} descriptor.json input_dir_or_json` |

### Tool Python Interfaces
```
In [1]: from boutiques.bosh import BoutiquesTools   # Load class from bosh
In [2]: bosh = BoutiquesTools()   # Create instance of tool class
In [3]: mode, desc, invoc = "launch", "descriptor.json", "inputs.json" 
In [3]: bosh.execute(mode, desc, invoc)
In [4]:   # Pass args the same as via the command-line-interface
In [4]:   # ... this is consistent for all tools
```

## Questions
(*ticked* means action has been taken)

### localExec
 - [x] Do we want to provide `--dontForcePathType` option to users, or just assume `False`?
    > verdict: remove option, assume `False`
 - [x] Do we want to provide `--changeUser` option to users, or just assume `True`?
    > verdict: remove option, assume `True`
 - [x] Do we want to change the `--keepTempScripts` flag to `--debug` and `-x` (the flag Unix uses for debugging)?
    > verdict: yes, make consistent with Unix
 - [ ] Should we update the randomly generated invocations to be generated in compliance with the schema?
    > verdict: yes, will do in a separate PR

### publish
 - [ ] Do we want to refactor this interface to take a descriptor and use the values therein, rather than a) requiring a descriptor lives in a git repo, and b) user to enter potentially redundant info?
   > verdict: yes, we want to simplify this and just require a descriptor be passed, but this requires https://github.com/boutiques/boutiques/issues/133 to be addressed


glatard:
Hi @gkiar, that's awesome! I think we do want the invocation functions to be exposed to the user. Platforms may need them, there is no real reason to hide them. 

About your questions: assume `False`, assume `True`, `yes`. 

Second-level `--help` flags seem to be crashing: ``$ ./bosh.py import --help``

Instead of having a third layer for `exec` (I assume `bosh exec simulate`), shouldn't we just have a separate command to generate random command lines? For instance `bosh examples`. By the way, we should really create invocations from these examples.

gkiar:
Great @glatard - integrated your answer to the above in the original comment. I do agree the examples should generate proper invocations, and will add that to the ToDo list (though perhaps a separate PR?)

I think it makes more sense to keep `simulate` and `launch` under exec - because it makes what I suspect to be a common use case more streamlined, i.e.:
 - 1 generate random inputs,
 - 2 create inputs I want,
 - 3 simulate execution with these inputs (i.e. print command-line), and
 - 4 run the tool

I think, in general, providing less choice to the user at each stage makes the user experience easier. So, instead of having 5 options at the top level and two of them closely related, having 4, and then in there having another choice between these 2 related functions, it will both implicitly form the link between these functions in the user's mind, and not cognitively burden them at any stage. Definitely want to hear your thoughts on the above.

gkiar:
Also, regarding the error you mentioned - there is no `def main()` or `if __name__ == "__main__"` condition in this file, so you should call it directly via the `console_script` entrypoint, i.e. `bosh import -h`. The command you used above would read the file, define the functions, but ultimately not do anything. I think this is cleaner because we don't actually need the `"__main__"` check to run our scripts as the package is currently configured.

gkiar:
@glatard I believe this is now ready for your review - see the initial comment updated for a description of both the Python and Command-line the interfaces, and any open questions (to be addressed either in this PR or others)

tools/python/boutiques/bosh.py
```diff
@@ -0,0 +1,280 @@
+#!/usr/bin/env python
+
+from argparse import ArgumentParser, RawTextHelpFormatter
+import jsonschema
+import json
+import os, sys
+
+class BoutiquesTools():
```
glatard: How about calling this `BoutiquesTool` (no plural)?

glatard: actually, how about having a module instead of a class? Currently, we need to create an object to access all the functions below, but this object has no attribute.

glatard:
I can't get the `--help` option to be working with `bosh exec launch`: 
```$ bosh exec launch -h
usage: Boutiques local executor [--help] {launch,simulate} descriptor
Boutiques local executor: error: too few arguments
```

glatard:
removing the `changeUser` option was not a good idea. Some tools, including `bids/example` wouldn't work with a new user.

gkiar:
Fixed the `help` interface in https://github.com/gkiar/boutiques/commit/6395a59a81ea8ecc8a9f4565fa220f48b2c766a7 - it required you to put the descriptor first, which now that I think about it was silly :)

tools/python/boutiques/bosh.py
gkiar: Conversation moved to your PR on this branch; https://github.com/gkiar/boutiques/pull/5

Issue #164, opened 2017-10-04
bosh: merge -r and -n


glatard:
`bosh exec simulate <tool.json> -r -n 5` could be written just `bosh exec simulate <tool.json> -r 5`.

PR #169, opened 2017-10-08
Merge -r and -n parameters for local executor's simulation mode

stefanches7:
#164


gkiar:
Hi @stefanches7 - thanks so much for this! I'd like to clarify the objective: we would like to remove the `-n` flag entirely and just have the number associated with `-r`. Meaning, there shouldn't be a need to add an import for the `SUPRESS` package or anything fancy like that, simply remove one command-line arg and update the interpretation of the other.

Does that make sense?

PR #173, opened 2017-10-31
Log handling and Docker test

ValHayot:
- bosh print statements returned as a variable (otherwise, unable to see statements if running on a Spark cluster). 
- added a skipif to docker test case so that it is skipped if docker is not installed

### DIFF
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -80,9 +80,9 @@ def execute(*params):
                                   "changeUser"         : results.user})
         executor.readInput(inp)
         # Execute it
-        exit_code, _, _ = executor.execute(results.volumes)
+        exit_code, _, err_dict = executor.execute(results.volumes)
         if exit_code:
-            raise SystemExit(exit_code)
+            raise SystemExit(exit_code, err_dict)
 
     if mode == "simulate":
         parser = ArgumentParser("Simulates an invocation.")
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -93,21 +93,18 @@ def execute(*params):
                             help="The Boutiques descriptor.")
         parser.add_argument("-i", "--input", action="store",
                             help="Input JSON complying to invocation.")
-        parser.add_argument("-r", "--random", action="store_true",
-                            help="Generate random set of inputs.")
-        parser.add_argument("-n", "--number", type=int, action="store",
-                            help="Number of random input sets to create.")
+        parser.add_argument("-r", "--random", action="store", type=int,
+                            nargs="*", help="Generate random set of inputs.")
         results = parser.parse_args(params)
         descriptor = results.descriptor
 
         # Do some basic input scrubbing
         inp = results.input
-        rand = results.random
-        numb = results.number
+        rand = results.random is not None
+        numb = results.random[0] if rand and len(results.random) > 0 else 1
+
         if numb and numb < 1:
             raise SystemExit("--number value must be positive.")
-        if numb and not rand:
-            raise SystemExit("--number value requires --random setting.")
         if rand and inp:
             raise SystemExit("--random setting and --input value cannot be used together.")
         if inp and not os.path.isfile(inp):
@@ -118,8 +115,6 @@ def execute(*params):
             raise SystemExit("JSON descriptor {} does not seem to exist.".format(descriptor))
         if not rand and not inp:
             raise SystemExit("The default mode requires an input (-i).")
-        if not numb:
-            numb = 1
 
         # Generate object that will perform the commands
         from boutiques.localExec import LocalExecutor
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -80,9 +80,9 @@ def execute(*params):
                                   "changeUser"         : results.user})
         executor.readInput(inp)
         # Execute it
-        exit_code, _, err_dict = executor.execute(results.volumes)
+        exit_code, _, err_msg = executor.execute(results.volumes)
         if exit_code:
-            raise SystemExit(exit_code, err_dict)
+            raise SystemExit(err_msg)
 
     if mode == "simulate":
         parser = ArgumentParser("Simulates an invocation.")
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -5,6 +5,7 @@ import jsonschema
 import json
 import os, sys
 
+
 def validate(*params):
     parser = ArgumentParser("Boutiques descriptor validator")
     parser.add_argument("descriptor", action="store",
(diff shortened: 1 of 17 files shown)


tools/python/boutiques/bosh.py
```diff
@@ -80,7 +80,7 @@ def execute(*params):
                                   "changeUser"         : results.user})
         executor.readInput(inp)
         # Execute it
-        exit_code = executor.execute(results.volumes)
+        exit_code, _, _ = executor.execute(results.volumes)
         if exit_code:
             raise SystemExit(exit_code)
```
glatard: how about printing the logs here?

tools/python/boutiques/localExec.py
```diff
@@ -122,15 +122,15 @@ def execute(self, mount_strings):
...
       print('Executing via: ' + dcmd)
-      exit_code = self._localExecute( dcmd )
+      (stdout, stderr), exit_code = self._localExecute( dcmd )
     # Otherwise, just run command locally
     else:
-      exit_code = self._localExecute( command )
+      (stdout, stderr), exit_code = self._localExecute( command )
     # Report exit status
-    print('---/* End program output */---\nCompleted execution (exit code: ' + str(exit_code) + ')')
+    log_data += '---/* End program output */---\nCompleted execution (exit code: ' + str(exit_code) + ')\n'
```
glatard: You don't seem to be doing anything with `log_data`, is it intentional?

gkiar:
I approve once the suggested changes are made :)

tools/python/boutiques/localExec.py
ValHayot: Actually, I wasn't sure what to do with it. I see value in having real-time print statements when working locally, but when running in distributed environment, I'd probably prefer having a log. So i don't know whether to keep both the print statements and the logs or just one of the two. If we do create a log, should it just be returned to the user, or should boutiques generate the file?

ValHayot: Naturally, this change shouldn't have been pushed. i thought i had reverted it back to the print statements, but it seems that with all my merges, the log came back.

ValHayot:
@gkiar Sorry about that! I decided to create a new comment entirely for now as I need advice (I'm unsure of all my changes). I'll update the original once i'm satisfied with what I've done.

- I've reverted back to the print statements. I don't know about returning the log as a variable. What if the function fails before returning the log? then there's just an error. However, I don't want to parse the slurm log to get the out of order print statements either. I'd like to be able to have the log for each node in a separate file. Is the answer to have both?
- errors are now returned as a dictionary, key 'Container' being that the script execution returns and key 'Tool' being the one matched to that indicated in descriptor (key names should be better). I thought of keeping it as a print statement, but I figured if we're returning the error to the user to deal with, might as well put it in an easier to parse format.
- the error dictionary is also included as the second argument of the SystemExit exception thrown by bosh (the first argument (arg[0]) being the error code)
- a test case has been added to validate that the correct tool error message was returned by SystemExit (another one should be created for stderr, but i need to find/create another tool as example1 writes stderr to file)

gkiar:
Hey @ValHayot thanks for the info! I think I understand, but to ensure I do, could you please paste examples of each of the above, what would be printed, and then the contents of the stored variables?

In general, I think putting everything to stdout would be great, and then we can usually find a way to interrupt stdout on systems and put them in log files for perusal.

ValHayot:
@gkiar sure for the stdout. i suppose i can also capture all the stdout from my application such that it doesn't even need to be added to Boutiques. Anyway, that has all been reverted, so it remains stdout!

- `_localExecute` returns a tuple produced by `process.communicate()` and the exit code (`process.returncode`). In other words: `(stdout, stderr), exit code`. 
In the case of example1 using `invocation_missing_script.json`, it would be `('', ''), 2`.  The reason for this is that all output is written to the file generated (log4.txt, or something). Otherwise, the  output would be something like `('some stdout', 'File noneexistentfile.py not found'), 2`
- In `[localExec.py] execute()`, a dictionary of max two error messages (could be more) is created. The first, being stderr message obtained from `_localExecute()` stored in key `Container`, and the second being the one produced from matching the error code to the tool descriptor. That's stored in key `Tool`.
For example1, it would be `{'Tool': 'File does not exist!'}`. If the stderr returned was not an empty string, it would have looked something like `{'Container': 'File noneexistentfile.py not found', 'Tool': 'File does not exist!'}. The assumptions I make are that there will only be one exit code produced and that it's possible to have both stderr and a descriptor-specified error at the same time. If this is not the case, then obviously a string makes more sense.
- `[localExec.py] execute()` returns the exit code, stdout and error dictionary 
- In `[bosh.py] execute()` the exit code and error dictionary are retrieved, and if there is an error, it will raise the `SystemExit` just as before, except the error dictionary will also be included (i.e. `SystemExit(exit code, error dict))`. For example1, that would be `SystemExit(2, {'Tool': 'File does not exist!'})`. The exit code will then be retrieved through `<SystemExit exeception instance>.args[0]` and the error dictionary will be retrievable through `<SystemExit exeception instance>.args[1]`.

Hope this helps!

glatard:
@gkiar we're ready to merge this when you approve it. Great work @ValHayot!

gkiar:
Hey @ValHayot thanks for the PR! I still had trouble groking the comments, but looking at the code I think simplified things a bit :) One open question: does this preserve all stdout and stderror, and show them alongside the error messages as well, since you're now capturing them? If so, can you show me an example stdout here just for "documentation"? Thanks! :D

ValHayot:
hey @gkiar sorry about that! changed things again. So, my previous comments don't apply at all. 
- If there's any stdout, it's printed right before the "completed execution part", that would be line 132
`if stdout != '': print('Execution output: '+ str(stdout))`
- Error messages (now in string format), are printed out by the SystemExit exception raised in bosh.py (line 85) or 
```
exit_code, _, err_msg = executor.execute(results.volumes)
if exit_code:
   raise SystemExit(err_msg)
``` 

and boutiques output will look something like this (sorry there is missing stdout, i think i may add another example other than example1, where stdout/stderr aren't appended to a file):
```
Attempting execution of command:
	exampleTool1.py -c ./config.txt -i foo bar    -e val1  ./nonexistentscript.py &>  log-4.txt
---/* Start program output */---
Container not found online - trying local copy
Executing via: docker run --entrypoint=/bin/sh --rm -e ENVAR='theValue'  -v ${PWD}:${PWD} -w ${PWD} boutiques/example1:test ./temp-1056571075214-1516338347548.localExec.boshjob.sh
---/* End program output */---
Completed execution (exit code: 2)
Looking for output files:
	Required output file 'Log file' was found at log-4.txt
	Optional output file 'outfiles' was not found at output/*_exampleOutputTag.resultType
	Required output file 'Configuration file' was found at ./config.txt
Example Boutiques Tool ERR (2): File does not exist!
```

gkiar:
Great, and one final question 😄 : if errors are unspecified in the descriptor, is the whole error message now dumped to the terminal as well? I.e. the substitution you showed applied if they specified error codes, but otherwise we now see full dump (previously, we saw something except error code)? Thanks!! Very excited about this. Also, can you please merge in the current develop branch? I just added a new PR

ValHayot:
Yep, stderr will be printed out as long as stderr is there. i'll take the opportunity to add another example!

gkiar:
@ValHayot Great! Are you still planning to add that example? All I'm waiting on is that and you to back-merge the develop branch (since your branch is a few commits behind).

PR #182, opened 2017-12-01
Develop

shots47s:
Changed the shell being used from /bin/bash to /bin/sh which is more universally supported by docker images, especially ones that are based on micro kernels.

Additionally, the typo I fixed in the tutorial is also a part of this PR.

addresses Issue #181


gkiar:
Seems great! @glatard - any additional thoughts? I'm happy to merge if you are

Issue #186, opened 2017-12-11
Tags


glatard:
Add an optional list of key-value pairs to describe the tool. No recommended schema for now. Discussed during the December meeting.

PR #194, opened 2018-01-10
Tags and arrays

glatard:
A tool descriptor may now contain a set of tags such as:
```
"tags": {
	"domain": "testing",
	"programming-language": "Python",
	"foo": "bar"
    }
```
It addresses #186. This is required for https://github.com/fli-iam/CARMIN/issues/79 

Besides, arrays `inputs`, `environment-variables`, `output-files` and `groups` cannot be empty and cannot contain duplicate elements.

### DIFF
--- a/tools/python/boutiques/schema/descriptor.schema.json
+++ b/tools/python/boutiques/schema/descriptor.schema.json
@@ -128,7 +128,9 @@
                     "value"
                 ],
                 "additionalProperties": false
-            }
+            },
+	    "minItems": 1,
+	    "uniqueItems": true
         },
         "groups": {
             "id": "http://github.com/boutiques/boutiques-schema/groups",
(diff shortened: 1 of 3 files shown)


tools/python/boutiques/schema/descriptor.schema.json
```diff
@@ -188,7 +190,9 @@
                     "members"
                 ],
                 "additionalProperties": false
-            }
+            },
+	    "minItems": 1,
```
gkiar: For groups, don't we want to allow the tool to have none of these?

glatard: yes we do, in this case there won't be any `groups` property. But if the property is there, the array can't be empty.

gkiar: Great, thanks! 😄

Issue #199, opened 2018-01-17
`bosh` exec should validate the descriptor before trying to run it


glatard:
User feedback (GigaScience reviewer):

When trying to run the example in https://github.com/boutiques/boutiques/tree/https://github.com/boutiques/boutiques/tree/5678d6f84bc1863741c6b8c595f037f622936c5c/tools/python/boutiques/schema/examples/example1/tools/python/boutiques/schema/examples/example1, validation of the invocation and descriptor functioned as described. For instance, validation failed when `enum_input` and `num_input`, mutually exclusive parameters, were both specified. Validation did not fail, however, when attempting to run a 'simulation' or 'launch' with the same invalid inputs. The authors should consider including the validation step when running the program with `bosh exec`.

gkiar:
Just addressed in a commit on my fork. Will close when we merge

Issue #198, opened 2018-01-17
Support for tools without a container image


glatard:
The spec allows for the definition of tools that don't have a container image. This is useful in case tools are pre-installed on the infrastructure (e.g. a cluster) or only simple shell utilities are used (e.g. for testing). We should make sure that `bosh` supports such tools and that this feature is covered by tests. 

User feedback (GigaScience reviewer):

I installed Boutiques version 0.5.3 (Python 3.6.3 on OS X 10.11.6) to test the minimal Boutiques descriptor in Listing 5. This descriptor passed the validation step, but when attempting to conduct a simulation run of the descriptor given an input JSON file with the contents `{"param": 5}`, a `KeyError` was produced, indicating that the key "container-image" was not present in the descriptor. This led me to wonder whether "container-image" is a required field in the descriptor schema and whether `bosh` can execute on bare metal.

gkiar:
This is now implemented in my branch - I will close this when we merge. The functionality existed, but the bug was in using `[]` access vs `.get()` access for dictionary properties (Ruby returns `None` in the former context, whereas Python throws an exception).

PR #200, opened 2018-01-17
Reviewer Feedback

gkiar:
- Addresses #198 
- Addresses #199

### DIFF
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -62,9 +62,11 @@ def execute(*params):
                             help="Runs the container as local user ({0}) instead of root.".format(os.getenv("USER")))
         results = parser.parse_args(params)
         descriptor = results.descriptor
+        inp = results.invocation
+
+        valid = invocation(descriptor, '-i', inp)
 
         # Do some basic input scrubbing
-        inp = results.invocation
         if not os.path.isfile(inp):
             raise SystemExit("Input file {} does not exist".format(inp))
         if not inp.endswith(".json"):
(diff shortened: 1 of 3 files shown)


tools/python/boutiques/bosh.py
```diff
@@ -62,16 +62,18 @@ def execute(*params):
...
-        inp = results.invocation
         if not os.path.isfile(inp):
             raise SystemExit("Input file {} does not exist".format(inp))
         if not inp.endswith(".json"):
             raise SystemExit("Input file {} must end in json".format(inp))
         if not os.path.isfile(descriptor):
             raise SystemExit("JSON descriptor {} does not exist".format(descriptor))
 
+        valid = invocation(descriptor, '-i', inp)
+        valid = invocation(descriptor, '-i', inp)
```
glatard: why is this line repeated twice?

glatard: shall we catch the exception and exit gracefully? Since it's a command-line tool and not the API?

gkiar: Typo, thanks :) 

I don't really understand what you mean in the latter point? For example, the last statement: scripts will normally exit if a file is not found.

glatard: if the descriptor is not valid, `invocation` will throw an exception. We may want to catch it and print a neat error message on stderr instead of throwing the exception at the user.

gkiar:
I added a test without a container (it's example 1 but slightly modified), and believe we already have tests for invalid descriptors, so wasn't sure what else you wanted me to add there.

tools/python/boutiques/bosh.py
gkiar: Ah, I understand. I've made invocation print in the same form as the validate module

glatard:
I guess that's enough tests (I was thinking we could test that the descriptor is validated before being executed, i.e., that execution of invalid decriptors fails with an "invalid descriptor" message rather than a runtime exception but I don't think it's necessary). However, your last commit didn't pass the tests :)

gkiar:
Just noticed that, which is annoying because the test passed locally! Updated anyways, and we'll see if that fixes things :)

PR #214, opened 2018-02-13
boutiques with tests

Trajan64:
# Descriptions and usage #

Boutiques will now be able to automatically perform a series of tests that are defined within a descriptor.

To try it out, simply type `bosh test <path to descriptor>` in a console.

Note that the provided descriptor must be augmented with the definition of a new property `tests`. 
See the following section to know how to define it

Also, note that:
- Sample JSON descriptors can be fond under `boutiques/tools/python/boutiques/schema/examples/` with names starting with tests.
- There is a test defined to automatically check the validity of this feature. It is located in `boutiques/tools/python/boutiques/test/ as test_test.py`
To perform those tests, a docker image must be mounted. It is the exact same image as the one needed for `test_exemple1.py`
You shall set it using the following command:
`docker build -t boutiques/example1:test ./tools/python/boutiques/schema/examples/example1`

## Structure of the `tests` property ##

The schema was augmented with the introduction of a new `tests` property. This property is an array which hold `test` objects that represent individual test that we wish to perform when using bosh test with the descriptor.

A test may be defined with the following items:
- A `name` (It has to be unique, no others tests may share the same name)
- An `invocation` object which represents the arguments for the pipeline program.
- An `assertion` object. It defines what pytest will have to check to figure out if the test passed or fail.
There are two items defined (at least one of those items has to be included in the descriptor) within an assertion object:
    - An `exit-code`, which is an integer representing the expected returned value by the pipeline.
    - An `output-files` item that represent an array of output-file objects, that have the following properties:
        - An id that refer to an output object. An `output` object, in this context, is an object held by the `output-files` array _which in turns, is defined at the top-level of the descriptor_
        - (Optional) A `reference` path that points to a file. The content of this file will be checked against the content of the file pointed to by the `id` property of the associated object.

In short, in regards to assertions, boutiques will do either, or both:
- Assert on the exit-code, if one was defined.
- Check each of the output-files defined in the array. More specifically, it will:
 (1) Check if the output file was indeed produced by the pipeline
 (2) If that is the case, and if a reference path was included, check if the output file's content matches the content of the file pointed by the reference path.

Also note that `name`, `invocation` and `assertion` are all required properties of a `test` object.

This is what the `tests` property looks like inside the JSON schema:

```
        [...]
        "tests": {
            "id": "http://github.com/boutiques/boutiques-schema/tests",
            "type": "array",
            "minItems": 1,
            "items": {
                "id": "http://github.com/boutiques/boutiques-schema/tests/test-case",
                "type": "object",
                "properties": {
                    "name": {
                        "id": "http://github.com/boutiques/boutiques-schema/tests/test-case/name",
                        "minLength": 1,
                        "description": "Name of the test-case",
                        "type": "string"
                    },
                    "invocation": {
                        "id": "http://github.com/boutiques/boutiques-schema/tests/test-case/invocation",
                        "type": "object"
                    },
                    "assertions": {
                        "id": "http://github.com/boutiques/boutiques-schema/tests/test-case/assertions",
                        "type": "object",
                        "properties": {
                            "exit-code": {
                                "id": "http://github.com/boutiques/boutiques-schema/tests/test-case/assertions/exit-code",
                                "description": "Expected code returned by the program.",
                                "type": "integer"
                            },
                            "output-files": {
                                "id": "http://github.com/boutiques/boutiques-schema/tests/test-case/assertions/out-files",
                                "type": "array",
                                "minItems": 1,
                                "items": {
                                    "id": "http://github.com/boutiques/boutiques-schema/test-case/assertions/out-files/output",
                                    "type": "object",
                                    "properties": {
                                        "id": {
                                            "id": "http://github.com/boutiques/boutiques-schema/tests/test-case/assertions/out-files/output/id",
                                            "description": "Id refering to an output-file",
                                            "minLength": 1,
                                            "pattern": "^[0-9,_,a-z,A-Z]*$",
                                            "type": "string"
                                        },
                                        "reference": {
                                            "id": "http://github.com/boutiques/boutiques-schema/tests/test-case/assertions/out-files/output/reference",
                                            "description": "Path to a file whose content will be compared against the outputed file matching the associated id",
                                            "minLength": 1,
                                            "type": "string"
                                        }
                                    },
                                "required": [
                                    "id"
                                ]
                                }
                          }
                        },
                        "anyOf": [
                          {
                            "required": [
                              "exit-code"
                            ]
                          },
                          {
                            "required": [
                              "output-files"
                            ]
                          }
                        ]
                    }
                },
                "required": [
                  "name",
                  "assertions",
		  "invocation"
                ]
            }
        },
        [...]
```

 
# Implementation details #

## The test engine ##

To allow a variable number of tests to be carried out by pytest, the new feature introduce 2 new files: `conftest.py` and `test.py`
`conftest.py` is a script that is meant to be executed by pytest before looking for tests to run (which in this case is just `test.py`).
The script forces pytest to call the test function within test.py, with different parameters every time.
Each call will provide enough information as parameters for the function, so that it can execute the pipeline and know how the subsequent results are expected to be asserted.

### `test.py` ###

`test.py` defines a function which, as stated in the section above, is called by pytest for each of the tests that are defined in the descriptor.
The function takes 3 parameters: A descriptor, the related test item from the descriptor and an invocation file.

The function first runs the pipeline with `bosh exec`, using both the descriptor and the invocation provided as parameters.
Subsequently, the function will then begin asserting.
It will first look if an exit-code was specified in the descriptor, and if such is the case, look at the captured exit code value to see if it matches that value inside the `test` JSON item provided as argument.
If `output-file` objects were provided in the descriptor, the function will iterate on each of them, and for each:
- ~~It will use a `LocalExecutor` instance to obtain a dictionary of output file ids mapped to actual file paths. **(NOTE: This is meant to go. A subsequent PR will use `bosh evaluate` instead)**.~~
- It will translate the ouput file id to an actual file path through a dictionary that was previously returned by a call to `bosh evaluate`.
- Assert on that the `output-file` object's `id` from the descriptor is found among the dictionary mentioned above.
- If a reference was specified, the function will:
    -- Assert on the actual existence of the file pointed by it.
    -- Compute the MD5 of both the output files and the reference files and assert on their equality.

### `bosh.py` ###

A method `test` was introduced to `bosh.py`. The function checks the descriptor provided as argument and prepare the call to `pytest`. In more details, the function will:
- Validate the descriptor and generate the invocation schema, all through one call to the invocation function.
- Generate the tests's invocations into temporary files and verify their validity through calls to the invocation function.
- Call `pytest` with the descriptor as parameter.
NOTE: There is room for improvement on that point. bosh.execute and bosh.invocation should be able to take objects as arguments and not just files. That would remove the need to rely on temporary files. Also, the process of generating temporary files for those same reasons is carried out in `conftest.py` too.

Additionally, bosh exec had its return statement modified. The function now returns a tuple: `(exit-code, stdout, stderr)`. It also print each of those elements onto `stdout` before returning.
Note also, that this applies for `bosh exec -launch`. `bosh exec -simulate` now only returns `(0, "", "")`.
Those modifications were made because being able to isolate the exit-code is a needed feature for `test.py`.

### `validation.py` ###
`validation.py` has been modified so as to include validating operations on the 'tests' property.
With the modifications made, there are 3 new possible errors that may arise during (the thorough) descriptor validation:
. Two or more tests share the same names.
. Two or more output-file (within one defined test) share the same output id.
. A test output-file refer to an invalid id (that is, there are no 'output' objects that have this id)

### ~~`localExec.py`~~ ###
~~Disregard the changes made to that file. I will use `bosh evaluate` instead.
Basically I needed a way to obtain a dictionary mapping output ids to file paths and I didn't think bosh provided this exact feature.~~
Changes applied; rolled back to original version.

## The tests (for this feature) ##
This new feature introduce a new testing file inside the `boutiques/tools/python/boutiques/tests/` directory. This file is (unoriginally) called `test_test.py`, and define 5 tests of different nature.
All those tests use the `exampleTool1.py` script, used in other previously defined tests such as `test_example1.py`.
In addition to retaking this sample pipeline, the script also retake the descriptors found in `boutiques/tools/python/boutiques/schema/examples/example1/example1.json`.
The JSON file remains the same, except that it is augmented with the new property 'tests', for which its definition is different for each of the (actual) tests.

- The first test is the only one that is meant to pass sucessfully. It defines 3 `test` objects, with various assertions strategies
(found in `boutiques/tools/python/boutiques/schema/example/tests_good.json`):
    - The first `test` object states that the assertions must be made on an exit-code of 0 and on the verification of the existence of a an output file. If it exists, the `test` object also state that an assertion must be made to check if its content matches the content of another file (`boutiques/tools/python/boutiques/schema/examples/ref_log.txt`)
    - The second `test` object specifies only an assertion on the exit-code (which is 0, again).
    - The third `test` object stipulates that the assertions must be made on the existence of two different output files (and no reference defined this time, so no content matching is required).

- The second test checks how bosh behave when given a descriptor with an invalid `tests` property (`boutiques/tools/python/boutiques/schema/examples/tests_invalid.json`)
It verifies that `bosh` correctly raises a `ValidationError`, and that the expected error messages are found in the exception's associated message.

- The third, fourth and fifth tests are all meant to fail during the assertions made by `test.py`.
Respectively, they test that an assertion failure happen when there is:
    - A mismatch between actual and expected exit-code (`boutiques/tools/python/boutiques/schema/examples/tests_failure_exitcode.json`)
    - A missing output-file (`boutiques/tools/python/boutiques/schema/examples/tests_failure_output_id.json`)
    - A MD5 mismatch between an output-file and its expected content (`boutiques/tools/python/boutiques/schema/examples/tests_failure_reference.json`)


gkiar:
Hi @Trajan64 - please look up the PR guidelines [here](https://github.com/boutiques/boutiques/blob/master/CONTRIBUTING.md#pull-request-process) and update the comment on your request. In particular, please describe the changes, and demonstrate your new feature.

Looking forward to this!

glatard:
* maybe also check why stdout and stderr are not correctly printed.

tools/python/boutiques/localExec.py
```diff
@@ -559,3 +559,10 @@ def check(keyname, isGood, msg, value): # Checks input values
       sys.stderr.write("Problems found with prospective input:\n")
       for err in self.errs: sys.stderr.write("\t" + err + "\n")
       sys.exit(1)
+
+  # Return a dictionary mapping output ids to computed output file paths
+  # Require an input file
+  def getOutputFiles(self, inp):
```
glatard: Check if you can use `bosh evaluate` instead of adding a new function.

Trajan64:
Hey @gkiar :), sorry about not including a comment. I decided to put the PR out there looking like this because I was about to have a meeting with @glatard. I wanted to try out the new feature directly with one of you guys before proceeding further.

Now that it's done, I'll update my comment ASAP to include a description on usage and the changes brought/implementation details.

Trajan64:
@gkiar, @glatard: I've edited my comment to include a full description of the new feature.
Please let me you know if you have any inquiries :)

glatard:
Hi @Trajan64, it looks good to me! No need to create a new PR for the requested changes, just add to this one: pushing to you master branch will add to it.

gkiar:
Hi @Trajan64 @glatard - I plan to look at this ASAP but the next few days are very busy for me; please don't merge until I review :) Excited to check it out!

tools/python/boutiques/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+    # Those informations are:
+    #                         . The descriptor (common to all)
+    #                         . The related JSON data, describing the test 
+    #                         (more convenient, no need to extract again from descriptor)
+    #                         . The invocation file needed for the test
+    descriptor_name, tests = fetch_tests(descriptor_filename[0])
+
+    # Generate the test ids for each of the test cases.
+    # An id is created by concatenaning the name of the descriptor with the name of the test case.
+    names = ['\'' + descriptor_name + "\'-\'" + parameters[1]["name"] + '\'' for parameters in tests]
```
gkiar: For readability, I'd write this as: 
```
"{}-{}".format(descriptor_name, params[1]["name"] for params in tests]
```

Out of curiosity, how come you added extra `'`s?

tools/python/boutiques/schema/descriptor.schema.json
```diff
@@ -384,6 +384,82 @@
...
+                              "output-files"
+                            ]
+                          }
+                        ]
+                    }
+                },
+                "required": [
+                  "name",
+                  "assertions",
+		  "invocation"
```
gkiar: missing 1 space here

tools/python/boutiques/schema/examples/tests.json
```diff
@@ -0,0 +1,473 @@
+{
```
gkiar: I'm a bit confused - how come this is a new file rather than just adding tests to the `good.json` file that we already had and which most of this was copied from?

tools/python/boutiques/test.py
```diff
@@ -0,0 +1,38 @@
...
+                
+                reference_path = output_file["reference"]
+                # Ensure that this property point to an existing file
+                assert os.path.exists(reference_path)
+                
+                # MD5 checksum comparaison
+                compute_md5 = lambda file_name: hashlib.md5(open(file_name,'rb').read()).hexdigest()
+                actual = compute_md5(file_path)
+                reference = compute_md5(reference_path)                
+                assert actual == reference
```
gkiar: use consistent formatting for assert - above you used brackets, subsequent 3 uses did not.

tools/python/boutiques/tests/test_example1.py
```diff
@@ -16,29 +16,29 @@ def test_example1_no_exec(self):
...
-                               os.path.join(example1_dir, "invocation.json")]))
+                               os.path.join(example1_dir, "invocation.json")])[0])
 
     @pytest.mark.skipif(subprocess.Popen("type docker", shell=True).wait(), reason="Docker not installed")
     def test_example1_exec_missing_script(self):
         example1_dir = os.path.join(self.get_examples_dir(), "example1")
-        with self.assertRaises(SystemExit) as context:
-           bosh(["exec", "launch", os.path.join(example1_dir, "example1.json"),
+        (exit_code, stdout, stderr) = bosh(["exec", "launch", os.path.join(example1_dir, "example1.json"),
                   os.path.join(example1_dir, "invocation_missing_script.json")])
```
gkiar: fix indentation to match line above (i.e. indent up to brackets)

tools/python/boutiques/tests/test_test.py
```diff
@@ -0,0 +1,49 @@
...
+class TestTest(TestCase):
+
+    def get_examples_dir(self):
+        return os.path.join(os.path.dirname(bfile),
+                            "schema", "examples")
+    
+    @pytest.mark.skipif(subprocess.Popen("type docker", shell=True).wait(), reason="Docker not installed")
+    def test_test_good(self):
+        self.assertFalse(bosh(["test",
+                     os.path.join(self.get_examples_dir(), "tests_good.json")]))
```
gkiar: fix indendation

tools/python/boutiques/tests/test_test.py
```diff
@@ -0,0 +1,49 @@
...
+from boutiques import bosh
+from jsonschema.exceptions import ValidationError
+
+class TestTest(TestCase):
+
+    def get_examples_dir(self):
+        return os.path.join(os.path.dirname(bfile),
+                            "schema", "examples")
+    
+    @pytest.mark.skipif(subprocess.Popen("type docker", shell=True).wait(), reason="Docker not installed")
```
gkiar: Just curious, why `type docker` over `which docker`? I don't think it matters, I've just always used `which` in this type of context :)

gkiar:
This looks great, just playing with it now! A bunch of comments and suggested minor modifications in line. In addition to the comments above, the following notes:

 - `pytest` should be added to the `DEPS` list in `setup.py`, so that without formally running tests or installing with `dev` mode, the function works.
- modify `cleanup_tests.sh` to checkout all modified files and remove extra ones. I think the following should do:
```
rm -r temp-*.sh log*.txt config*.txt .pytest_cache/
git checkout boutiques/schema/examples/
```

tools/python/boutiques/bosh.py
```diff
@@ -81,9 +82,11 @@ def execute(*params):
...
         executor.readInput(inp)
         # Execute it
-        exit_code, _, err_msg = executor.execute(results.volumes)
-        if exit_code:
-            raise SystemExit(err_msg)
+        exit_code, stdout, stderr = executor.execute(results.volumes)
+        print("Exit code: {}".format(exit_code))
+        print("Standard output: \"{}\"".format(stdout))
+        print("Standard error: \"{}\"".format(stderr))
+        return (exit_code, stdout, stderr)
```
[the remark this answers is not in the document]
Trajan64: I believe it's better to return all 3. 
As an example it could be useful to capture the pipeline's output.

However, in the context of bosh test, only the exit code is needed.

tools/python/boutiques/bosh.py
```diff
@@ -277,6 +281,37 @@ def evaluate(*params):
...
+    result = parser.parse_args(params)
+
+    # Generation of the invocation schema (and descriptor validation).
+    invocation(result.descriptor)
+    
+    # Extraction of all the invocations defined for the test-cases.
+    descriptor = json.loads(open(result.descriptor).read())
+    
+    if (not descriptor.get("tests")):
+        raise SystemExit("No test found in descriptor")
```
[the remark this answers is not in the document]
Trajan64: Yes that's a good point.

I'll make the modifications so that bosh test simply return 0 in this case.
0 is the exit code returned by pytest if all the tests have passed successfully.

tools/python/boutiques/bosh.py
```diff
@@ -277,6 +281,37 @@ def evaluate(*params):
...
+    
+    for test in descriptor["tests"]:
+        # Create temporary file for the invocation() function.
+        invocation_JSON = test["invocation"];
+        temp_invocation_JSON = tempfile.NamedTemporaryFile(suffix=".json", delete=False)
+        temp_invocation_JSON.write(json.dumps(invocation_JSON).encode())
+        temp_invocation_JSON.seek(0)
```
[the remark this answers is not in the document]
Trajan64: Yes that's something that I have highlighted in the description above.

Modifying bosh so that at least bosh invocation can take objects as invocations would make the code in both bosh test and conftest.py much cleaner. It would also remove the need to rely on intermediate files just like you have stated.

So yes I agree :) 

I think for all the bosh functions, it is simply a matter of adding an if case that checks if the invocation/descriptor parameter is a variable of type dict. In which case bosh will simply skip over the part about opening and loading up the JSON.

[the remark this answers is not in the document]
gkiar: Rather than writing it each time, let's write a little utility and call it in our functions?

tools/python/boutiques/bosh.py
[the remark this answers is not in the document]
gkiar: Great, I agree :)

tools/python/boutiques/conftest.py
Trajan64: I added extra ' because the name of the descriptor can take whitespaces and any other characters. If I run one of the JSON file which include failing test cases, pytest shows me "Example Boutique Tool-test1" (or something just like that). I thought including ' would better delimit the name of the descriptor and the test case.

I actually wasn't so confident with this idea. Feel free to share a suggestion on that.

tools/python/boutiques/bosh.py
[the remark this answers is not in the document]
glatard: I agree, but can we merge this first and then discuss how the API should be modified to support objeects?

tools/python/boutiques/tests/test_test.py
glatard: I think @ValHayot introduced that in other parts of the tests. I had checked the exact same thing at that time: `type` is a bash built-in while `which` isn't, which makes the code more robust in case `which` is not installed. That said, it might crash with `sh`.

ValHayot: Yeah, i had searched through stackoverflow to find the best command. Apparently missed `command -v` though, which would have us covered for everything: https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script

tools/python/boutiques/schema/examples/tests.json
Trajan64: Yes that's true. I can c/p the tests in tests_good.json (that are meant to pass successfully) to good.json

However know that the other test_failure* JSON files are technically correct, they just fail because the test cases do not pass. Do you still want to proceed with that information ?

BTW disregard tests.json, this file should have been deleted.

gkiar: OK perfect, the only one I didn't love was `tests.json` :)

tools/python/boutiques/bosh.py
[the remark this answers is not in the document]
gkiar: I'm happy merging first, but I don't think the API will change at all - each routine will just evaluate the object it receives (dict or file), and will return the dict version of that data.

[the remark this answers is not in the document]
glatard: okay then feel free to do it if it's clear in your mind @Trajan64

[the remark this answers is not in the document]
Trajan64: Yes sure :) The modifications should be very light.
And I agree with @gkiar's perspective, I'll try to implement a small function to perform the task.

tools/python/boutiques/conftest.py
gkiar: OK, so how about you replace spaces with `-` and separate descriptor name with a `_`? I.e:

```
"{}_{}".format(descriptor_name, params[1]["name"].replace(' ', '-') for params in tests]
```

Trajan64: Are you sure you want to modify the name of the descriptor ? I personally think it's better not to alter it.
However what could be done is that we could modify the schema so as to disallow white-spaces and `_`, but that might be overkill for such a small concern.

Another idea I had was to format it like that: `desc:"Exemple Test Tool"; test:"test1"`. But I thought not to go that way because it doesn't really feel much like an ID anymore (which might have appliances other than just displaying tests that failed).

gkiar: @Trajan64 I'm not sure I understand... With what you proposed, a descriptor named "`mydesc.json`", and two tests, "`my test 1`" and "`my test 2`", we'd get:
```
'mydesc.json'-'my test 1'
'mydesc.json'-'my test 2'
```

With my proposal, we'd get:
```
mydesc.json_my-test-1
mydesc.json_my-test-2
```

I prefer mine simply because it has no spaces or special characters in it. What's the issue?

Trajan64: Oh okay, I think we have a confusion on the matter of the descriptor name.
For me the descriptor name is the name object inside the descriptor, and not the filename.
For `tests_good.json`, there is the following property:  `"name": "Example Boutiques Tool"`
So as a consequence, with what I originally proposed and `tests_good.json` we have `'Exemple Test Tool'-'test1'` for one of the tests.

And also, I read your code too fast. I was under the impression that you were altering the white spaces within the descriptor name (as in the name object in this case).

Anyway, if you prefer we can use the filename instead. In which case I agree with your proposed solution.

gkiar: Ah! Ok, ya I think let's stick to the filename rather than descriptor object name, since I agree it'd be nice to not change the descriptor name but I don't want to have spaces in an ID :)

What do you think?

Trajan64: I agree. It's true its not ideal at all to have white-spaces in an ID.

So good with me :) I'll make the changes as you stated, using your piece of code.

gkiar: Great!!!

tools/python/boutiques/bosh.py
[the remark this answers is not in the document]
Trajan64: Turns out there's no clean way to handle this issue.
It is apparently impossible to allow `argparse` to accept dictionary objects as arguments. 
Which makes sense, but I wanted to enable the usage of the inner bosh functions directly with dictionary objects.
And as `argparse` is included in the beginning of every functions, this is just impossible to do.

There are multiple alternatives that can be used as work-around, but none of them are clean or simple:
- (1) Allow representation of dictionaries as raw `str` objects. This is tricky as each functions will have to differentiate between a string representing a dictionary, and a string representing a file path.
  - One way to perform this differentiation is to check if the first and last characters of the `str` object are respectively `{` and `}`
  - Another is to setup a try except block with a `json.loads` (loading for raw data) and if it works, we can be sure its a proper dictionary represented as a string and not a file path.
- (2) I think I can bypass `argparse` behavior by setting up a list containing of a string as first element, and the proper JSON dictionary as second element
- (3) Put all the `argparse` logic in the bosh function.

Also note that the modifications required go beyond `bosh.py`: `localExec.py` (`__init__` and `readInput` functions) and `validate.py` (`__init__` only) need to be modified so that they can choose to open a file with the provided object as string path, or read directly from the object to get the JSON.

For now I'll leave the problem as it is and will await your directions on the matter.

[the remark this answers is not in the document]
gkiar: @Trajan64 no problem - you don't need to implement this in this PR. I have done something similar to this before, so will take a crack at it myself later on. I suspect your assessment is correct, but that there are simplification we can make to still have this work for us (i.e. wrap it as a string as you suggest, etc.)..

I mentioned to @glatard that I've made plans to overhaul the API organization a bit again with some recent lessons learned, and will do this then. So as @glatard suggested upstream here, please ignore this suggestion in the PR, and let us know when all other suggested changes are done.

Cheers!

Trajan64:
Hello @gkiar and @glatard. All the suggestions have now been applied.
Let me know your thoughts :)

tools/python/boutiques/schema/examples/tests.json
gkiar: @Trajan64 should this file, `tests.json` still have been deleted?

gkiar:
Looks great - I have one remaining question about the deletion of `tests.json`, but otherwise I think it's all set! Thanks, @Trajan64 !!! 😄

gkiar:
Oh, and also, did you do the edits mentioned in my final comment? Updating `setup.py` and the `cleanup_tests.sh` scripts?

glatard:
This looks awesome @Trajan64, great work! maybe fix the final small edits suggested by @gkiar above and then we merge?

Trajan64:
@glatard Yes you're right, that would be better.
Let's hold the merge off for now until this last feature/fix is introduced. :)

Trajan64:
And thank you @gkiar for mentioning again about `tests.json`, `setup.py` and `cleanup_tests.sh`.
I had forgotten about those 3 files.

Now `tests.json` has been deleted and both `setup.py` and `cleanup_tests.sh` have been updated. :)

glatard:
@shots47s made a great suggestion yesterday: instead of storing the reference file in the container, just add its md5 to the descriptor (in place of 'reference' I guess). This way, you don't have to copy the reference out of the container: just compute the md5 of the result file and compare it to the value in the descriptor. Sounds good?

Trajan64:
@glatard : Yes, this is a good idea. :)

So if I understand correctly, the JSON schema would have to be changed so that the `reference` property would not describe a file path but a string containing a MD5 checksum (And probably change the name of the property to `reference-MD5` or something of the sort)

This also implies that the user will have to calculate the MD5 on his own and set it up himself in the descriptor, is that correct ?
That might be the downside of the strategy however, but it's true that it would make the issue much more simpler. I could probably add this change to the PR very quickly.

glatard:
Yes, you got it right. Let's start like this and see later how to improve based on a first version.

Trajan64:
I've replaced the md5 computation with the direct inclusion of the md5 within the descriptor itself, as you have requested.
I believe that the PR is ready to be merged :)

glatard:
This is all good for me. @gkiar, do you confirm?

gkiar:
Looks great @Trajan64 @glatard - thanks! :)

PR #221, opened 2018-03-12
Exporter

glatard:
Adds an 'export' command to bosh to export descriptors to CARMIN format. 

Needed by @louis-ver and @simondube's project.

### DIFF
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -4,8 +4,7 @@ from argparse import ArgumentParser, RawTextHelpFormatter
 import jsonschema
 import json
 import os, sys
-import tempfile
-import pytest
+
 
 def validate(*params):
     parser = ArgumentParser("Boutiques descriptor validator")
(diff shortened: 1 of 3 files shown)


tools/python/boutiques/bosh.py
```diff
@@ -148,7 +147,29 @@ def importer(*params):
...
+    parser.add_argument("type", help="Type of export we are performing. For carmin, pipeline id is set to the output file name.",
+                        choices=["carmin"])
+    parser.add_argument("descriptor", help="Boutiques descriptor to export.")
+    parser.add_argument("output", help="Output file where to write the converted descriptor.")
+    results = parser.parse_args(params)
+
+    descriptor = results.descriptor
+    output = results.output
 
+    try:
```
gkiar: Why not use the `bosh.validate()` here?

tools/python/boutiques/tests/test_export.py
```diff
@@ -0,0 +1,19 @@
...
+from boutiques import bosh
+from boutiques import __file__ as bfile
+
+
+class TestImport(TestCase):
+    def get_examples_dir(self):
+        return os.path.join(os.path.dirname(bfile),
+                            "schema", "examples")
+
+    def test_export(self):
```
gkiar: Do we have an example for what this descriptor should look like in CARMIN format? If so, we should also have an assertion that they are equal. No need to hold up the PR for this, but comment if you agree and we'll make an issue.

glatard: We don't at the moment. I haven't checked that this passes CARMIN validation. Counting on @louis-ver and @simon-dube to clarify that. See https://github.com/boutiques/boutiques/issues/222

tools/python/boutiques/bosh.py
glatard: good point.

PR #225, opened 2018-03-17
Fixing "invocation" typo

simon-dube:
Fixing a small "invocation" typo in print


glatard:
Perfect, thanks @simon-dube!

PR #227, opened 2018-03-19
Modifications for Singularity

gkiar:
- updated `bosh` to return the exit code properly via the CLI
- now supports multiple indices for `container-image` in the context of `singularity`
- will properly download and rename images pulled through `singularity`, and won't download image if there is one in the working directory with the same name

This PR is necessary for `bosh` to work using Singularity on a cluster (tested on Compute Canada, Cedar). Post merging this, given another PR with small modifications, I will release `0.5.6-post1` on pypi.

### DIFF
--- a/tools/python/boutiques/localExec.py
+++ b/tools/python/boutiques/localExec.py
@@ -85,17 +85,21 @@ class LocalExecutor(object):
     if conIsPresent:
       if conType == 'docker':
         # Pull the docker image
-        if self._localExecute("docker pull " + str(conImage)):
+        if self._localExecute("docker pull " + str(conImage))[1]:
           print("Container not found online - trying local copy")
       elif conType == 'singularity':
         if not conIndex:
           conIndex = "shub://"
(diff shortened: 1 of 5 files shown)


tools/python/boutiques/localExec.py
```diff
@@ -80,16 +80,26 @@ def execute(self, mount_strings):
...
+        if not conIndex:
+          conIndex = "shub://"
+        elif not conIndex.endswith("://"):
+          conIndex = conIndex + "://"
+        conName = conImage.replace("/", "-") + ".simg"
+
+        if conName not in os.listdir('./'):
+          print(os.listdir('./'))
+          # Pull the singularity image
+          if self._localExecute("singularity pull --name \"{}\" {}{}".format(conName, conIndex, conImage))[1]:
```
glatard: It would be nice if this could print the command that is being executed as bosh would typically hang on that for a while. Maybe add a print in localExecute?

Issue #193, opened 2018-01-07
Very cryptic error message provided when input.json has input parameters not in line with boutiques descriptor.


shots47s:
We should add some error handling that provides the users with more relavant information for debugging as now if you have something in the input.json that is not in line with the boutiques descriptor, the error that is sent back looks like this:

File "python3.5/site-packages/boutiques-0.5.4-py3.5.egg/boutiques/localExec.py", line 39, in <lambda>
    self.byId       = lambda n: [ v for v in self.inputs+self.outputs if v['id']==n][0]
IndexError: list index out of range

Which is not very helpful.  

Propose to have a input file validation method be created that checks this and fails with meaningful error messages.

gkiar:
Thanks, @shots47s! I agree, that would be better :)

PR #239, opened 2018-04-10
JSONLint and minor style fixes

glatard:
* Added JSONLint of JSON schema.
* `bosh` now returns validation exception when called from a Python script, but only prints an error message when called from the command-line (fix for #193). 
* Minor style fixes


gkiar:
@glatard seems great!!

PR #243, opened 2018-04-17
Added more tests

glatard:
* tested the bosh CLI instead of just the API
* cleaned-up dead code
* tested auth error in Zenodo


gkiar:
Closing - I just merged this into #244

Issue #172, opened 2017-10-26
bosh exec crashes if `container-image` is not defined.


glatard:
We should either:
* make `container-image` mandatory in the schema, or:
* make bosh exec able to handle descriptors that don't have a `container-image`.

glatard:
It was already fixed, added a test in #246.

PR #246, opened 2018-04-18
Small fixes

glatard:
* Added a test for tools with no container image (#172)
* `bosh` CLI now returns `99` in case an exception occurs, to reduce potential conflict with tool exit code.
* Added a test for invalid JSON descriptors and caught the exception in validator.
* Raised an exception when mandatory output file is not found (was just printing a message).

### DIFF
--- a/tools/python/boutiques/tests/test_no_container.py
+++ b/tools/python/boutiques/tests/test_no_container.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python
+
+import os
+import subprocess
+import pytest
+from unittest import TestCase
+from boutiques import __file__ as bfile
+import boutiques as bosh
+
+
+class TestExample1(TestCase):
+
+    def get_examples_dir(self):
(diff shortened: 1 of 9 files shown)


tools/python/boutiques/tests/test_no_container.py
```diff
@@ -0,0 +1,23 @@
...
+        return os.path.join(os.path.dirname(bfile),
+                            "schema", "examples")
+
+    def test_no_container(self):
+        self.assertFalse(bosh.execute("launch",
+                                      os.path.join(self.get_examples_dir(),
+                                                   "no_container.json"),
+                                      os.path.join(self.get_examples_dir(),
+                                                   "no_container_invocation."
+                                                   + "json"))[2])
```
gkiar: not that it matters, but I don't think we need the `+`

gkiar:
Looks great! Merge when ready :) You're on a roll!!!!

PR #251, opened 2018-04-19
WIP: Adding singularity testing to boutiques

gkiar:
(do not merge yet)

Changes:
- updated `.travis.yml` to include Singularity setup

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
-language: python
-
+dist: trusty
 sudo: required
+language: python
 
 services:
 - docker
@@ -19,9 +19,24 @@ python:
 node_js:
 - 5
 
+env:
+- SINGVER=2.4.6
+
 before_install:
-- docker build -t boutiques/example1:test ./tools/python/boutiques/schema/examples/example1
+- sudo apt-get update -qq
+- sudo apt-get install -y -qq python dh-autoreconf squashfs-tools build-essential
 - npm install jsonlint -g 
+- docker build -t boutiques/example1:test ./tools/python/boutiques/schema/examples/example1
+- docker run -v /var/run/docker.sock:/var/run/docker.sock -v ${HOME}:/output --privileged -t --rm singularityware/docker2singularity boutiques/example1:test  1> ${HOME}/convert_log.txt
+- IMGNAME=$(cat ${HOME}/convert_log.txt  | grep "Done. Image can be found at:" | awk '{ print $7 }' | cut -d / -f 3 | cut -d . -f 1)
+- wget https://github.com/singularityware/singularity/releases/download/$VER/singularity-$SINGVER.tar.gz
+- tar xvf singularity-$SINGVER.tar.gz
+- cd singularity-$SINGVER
+- ./configure --prefix=/usr/local --sysconfdir=/etc
+- make
+- sudo make install
+- singularity --version
+- cd ${HOME}
 
 install:
 - pip install --upgrade setuptools
@@ -30,6 +45,7 @@ install:
 
 script:
 - cd ./tools/python/
+- mv ${HOME}/${IMGNAME}.img ./boutiques-example1:test.simg
 - pycodestyle --max-line-length=80 boutiques/*.py boutiques/tests/*.py
 - jsonlint -q boutiques/schema/descriptor.schema.json
 - coverage run --source boutiques setup.py pytest


tools/python/boutiques/localExec.py
```diff
@@ -206,22 +206,52 @@ def execute(self, mount_strings):
...
+                dcmd = (envString + 'singularity exec --cleanenv ' +
                         singularity_mounts + ' -W ' + launchDir + ' ' +
                         str(conName) + ' ' + dsname)
             else:
                 print('Unrecognized container type: \"%s\"' % conType)
                 sys.exit(1)
             print('Executing via: ' + dcmd)
             (stdout, stderr), exit_code = self._localExecute(dcmd)
+            print(stdout)
+            print(stderr)
```
glatard: I don't think we want to do this here, in case this is called from the API. We should print only in `bosh`, when executing through the CLI.

tools/python/boutiques/localExec.py
```diff
@@ -206,22 +206,52 @@ def execute(self, mount_strings):
...
+                #
+                # Previous bind string:
+                #   singularity_mounts = " -B ".join(m for m in mount_strings)
+
+                def_mounts = ["/proc", "/sys", "/dev", "/tmp", "/var/tmp",
+                              "/etc/localtime", "/etc/hosts",
+                              op.realpath(op.expanduser('~')),
+                              op.expanduser('~')]
+                singularity_mounts = ""
+                for m in mount_strings:
```
glatard: What if an `m` in `mount_strings` is prefix of another `n` in `mount_string`?

tools/python/boutiques/localExec.py
```diff
@@ -206,22 +206,52 @@ def execute(self, mount_strings):
...
+                    if not any(d in m for d in def_mounts):
+                        singularity_mounts += "-B {0} ".format(m)
+                print(mount_strings)
+                print(singularity_mounts)
```
glatard: Prints should be done only in `bosh`, when using the CLI. We might want to add debug info to the tuple returned by this method if this is useful to the API. Or do this only in debug mode.

tools/python/boutiques/tests/test_example1.py
```diff
@@ -14,27 +14,52 @@ def get_examples_dir(self):
...
+                                                   "example1_docker.json"),
                                       "-i",
                                       os.path.join(example1_dir,
                                                    "invocation.json"))[2])
 
     @pytest.mark.skipif(subprocess.Popen("type docker", shell=True).wait(),
                         reason="Docker not installed")
-    def test_example1_exec(self):
+    def test_example1_exec_docker(self):
+        example1_dir = os.path.join(self.get_examples_dir(), "example1")
```
glatard: How about cleaning up here too, just in case?

tools/python/boutiques/tests/test_example1.py
```diff
@@ -46,7 +71,7 @@ def test_example1_exec_missing_script(self):
         sout, serr, ecode, emsg = bosh.execute(
```
glatard: add clean_up before that?

tools/python/boutiques/tests/test_example1.py
```diff
@@ -14,27 +14,52 @@ def get_examples_dir(self):
...
                         reason="Docker not installed")
-    def test_example1_exec(self):
+    def test_example1_exec_docker(self):
+        example1_dir = os.path.join(self.get_examples_dir(), "example1")
+        self.assertFalse(bosh.execute("launch",
+                                      os.path.join(example1_dir,
+                                                   "example1_docker.json"),
+                                      os.path.join(example1_dir,
+                                                   "invocation.json"))[2])
+        self.assertFalse(bosh.execute("launch",
```
glatard: how about cleaning up here too?

tools/python/boutiques/validator.py
```diff
@@ -123,10 +123,11 @@ def inById(i):
             if clkeys[jdx] == key:
                 mids = [inById(mid)["id"] for mid in inIds
                         if inById(mid)["value-key"] == key]
-                for idx, grp in enumerate(descriptor["groups"]):
-                    mutex = grp.get("mutually-exclusive")
-                    if set(grp["members"]) == set(mids) and not mutex:
-                        errors += [msg_template.format(key)]
+                if len(grpIds) > 0:
+                    for idx, grp in enumerate(descriptor["groups"]):
```
glatard: How about simply `enumerate(descriptor.get("groups"))`?

PR #257, opened 2018-04-23
Identifiers in CARMIN exports

### LINKED ISSUE #256: In CARMIN exporter, make ids non random
So that the CARMIN-server could recompute them safely at every reboot.

glatard:
This fixes #256. I realize that we may want to have a discussion on what should be a pipeline identifier but @simon-dube and @louis-ver need this a.s.a.p to complete the CARMIN server. This PR provides a flexible solution where the identifier is set by the user.


gkiar:
I still think there should be a conditional that if no ID is passed and there is a doi, use the doi. That doesn't make an issue for debugging if you do specify an ID, but gives actually meaningful IDs in production

PR #258, opened 2018-04-24
Improved tests and update to bosh exec launch

### LINKED ISSUE #160: Improve testing
* Most tests currently only check that no error is raised. The correctness of results should also be checked.
* Integration tests should be written, combining existing `bosh` commands wherever appropriate.

glatard:
* Improved various tests to check correctness of output instead of just absence of error (fixes #160).
* Updated `bosh exec` so that it returns an object instead of a long tuple. Object contains:
    *  `stdout`
    * `stderr`
    * `exit_code`
    * `error_message`: the error message associated with `exit_code` in the descriptor
    * `missing_files`: an array of missing files
    * `output_files`: an array containing the files produced by the execution
    * `shell_command`: the shell command executed in the container
    * `container_command`: the command used to start the container
    * `container_location`: the location of the container used (local or pulled)
* bosh CLI prints the return object (see screenshot below) and returns the exit code, bosh API just returns the object.
* Replaced `sys.exit` in `bosh exec` by exceptions, so that the API can catch them.
* Removed spurious prints in `bosh exec`.
* Removed various spurious license headers.

[screenshot from 2018-04-24 12-57-45](https://user-images.githubusercontent.com/5174953/39202166-22a0eed0-47bf-11e8-877c-d0c2fef07fdc.png)


tools/python/boutiques/localExec.py
```diff
@@ -278,41 +358,39 @@ def execute(self, mount_strings):
                 os.remove(dsname)
         # Check for output files (note: the path-template
         # can contain value-keys)
-        print('Looking for output files:')
+        missing_files = []
+        output_files = []
         for outfile in self.desc_dict['output-files']:
```
gkiar: I think we can use `evaluate` for a lot of this loop, no? I don't remember all of the functionality built in to it, but it is able to at least query outputs for optional vs required, breaking this into a couple smaller loops. If you want I can implement this and merge into your fork.

tools/python/boutiques/tests/test_import.py
```diff
@@ -4,17 +4,25 @@
...
 import json
 import os
+import os.path as op
+from os.path import join as opj
 
 
 class TestImport(TestCase):
 
     def test_import_bids_good(self):
         bids_app = op.join(op.split(bfile)[0],
```
gkiar: Since you're using `opj` many places in this file, might as well use it everywhere we use `op.join`? Can't remove the `import os.path as op` anyways since we use other functions of it, but consistency is nice :)

gkiar:
Awesome, thanks for this!

Issue #175, opened 2017-11-07
Escape special characters in command-line substitutions


glatard:
Question from @gflandin: who should escape special characters on the command line?

Maybe we could enforce that in `bosh`, by single-quoting all substituted values.

glatard:
Proposal from the meeting: single-quote all substituted values

jbpoline:
would shlex.quote()  be the python 3 tool for this ?
https://stackoverflow.com/questions/35817/how-to-escape-os-system-calls-in-python

glatard:
`shlex.quote` only works in Python 3. Maybe we can fall back on `pipes.quote` when `shlex.quote` isn't there, as suggested here: https://stackoverflow.com/questions/26790916/python-3-backward-compatability-shlex-quote-vs-pipes-quote
```
try:
    from shlex import quote as cmd_quote
except ImportError:
    from pipes import quote as cmd_quote
```

ramou:
When passing a single-quote-encapsulated string in my invocation I had some grief. The shell command told me it was:
```
blast_formatter  -archive out.asn -outfmt '7 qacc sacc evalue qstart qend sstart send'
```
which runs fine live, but gave an error message about formatted arguments from bosh (invoked programatically, though I think that's not an issue). The input of interest is `outfmt`

The invocation was:
```
		"invocation": {
			"archive": "out.asn",
			"outfmt": "'7 qacc sacc evalue qstart qend sstart send'",
			"out": "blast_formatter.out"
		},
```
You can invoke this through the tests (which fail do to the misinterpretation of the quotes), but I had trouble invoking through the command-line as I had to escape the single quotes and that messed stuff up. You can invoke the other variants through the command-line:

```
bosh exec launch ../blast_formatter.json '{"archive": "out.asn", "outfmt": "7 qacc sacc evalue qstart qend sstart send", "out": "blast_formatter.out"}'
```
does not work and yields the same _shell command_ as the broken test invocation above:
```
bin/blast_formatter  -archive out.asn -outfmt '7 qacc sacc evalue qstart qend sstart send'       -out blast_formatter.out
```

On the other hand, the following (which is what I changed the test invocation to in order to make it work) gives the desired result:

```
bosh exec launch ../blast_formatter.json '{"archive": "out.asn", "outfmt": "\"7 qacc sacc evalue qstart qend sstart send\"", "out": "blast_formatter.out"}'
```
works and yields the _shell command_:
```
[...]

PR #259, opened 2018-04-24
Custom shell and string escapes

### LINKED ISSUE #89: In LocalExec, string and file inputs should be single quoted

### LINKED ISSUE #188: Allow tools to use a custom shell
`bosh` currently uses `/bin/sh` to run the command line. The shell could be defined by the tool, and retrieved through the SHELL environment variable in the container. Discussed at the December meeting.

glatard:
* Fixed #188. Added a `shell` property to the descriptor that defines the absolute path of the shell interpreter to use in the container.
* Fixed #89 and #175. All string or file inputs are escaped before being inserted on the command line or in a configuration file. Other input types are not escaped. Output file names are not escaped: a file named `a;b.txt` can be produced but it will appear as `'a;b.txt'` on the command line.

### DIFF
--- a/tools/python/boutiques/test.py
+++ b/tools/python/boutiques/test.py
@@ -15,6 +15,7 @@ def test(descriptor, test, invocation):
     ret = bosh.execute("launch",
                        descriptor,
                        invocation.name)
+    print(ret)
 
     # Choose appropriate assertion scenario
     assertions = test["assertions"]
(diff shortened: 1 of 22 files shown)


tools/python/boutiques/test.py
```diff
@@ -15,6 +15,7 @@ def test(descriptor, test, invocation):
     ret = bosh.execute("launch",
                        descriptor,
                        invocation.name)
+    print(ret)
```
gkiar: Do we want this print statement, or was it for debugging?

glatard: Both :smiley: It is useful to have a more verbose stdout in tests when they fail. In case it's unclear, here we're talking about tool tests launched by `bosh test`.

PR #270, opened 2018-05-23
Bug fixes

glatard:
* Fixed tests for Python 2.6: pyOpenSSL, a dependency of coveralls, dropped support for Python 2.6 in May, which created side effects in web requests used by the publisher. We now install an older version of pyOpenSSL for coveralls when using 2.6 on Travis. This has no impact outside of a Travis install.
* Fixed exception message print in Python 3 and added a test for it.
* Fixed conflict between master and develop (not sure how it had happened, probably due to my PR from glatard:develop to master)
* Added better stdout messages in CLI mode for functions returning None.

Will release 0.5.12 when this is approved.


tools/python/boutiques/bosh.py
```diff
@@ -420,7 +425,7 @@ def bosh_return(val, code=0):
         if runs_as_cli():
             try:
                 print(e.message)
-            except Exception as e:
+            except Exception as ex:
                 print(e)
```
gkiar: need to change to `print(ex)` I think, since you renamed the exception from `e` to `ex`?

glatard: that was exactly the bug: we want `e` printed, not `ex`. In Python 3, `ex` would just be `object has no attribute message`.

PR #272, opened 2018-05-28
Improved console output

glatard:
* Added more verbose messages for Singularity containers
* In localExec, `-x` now prints debug messages when it executes a command

Apologies for the multiple commits, let's squash and merge all that!


glatard:
okay @gkiar, it took me an insane amount of time but this is now passing the tests with increased coverage... ready for another review!

PR #275, opened 2018-07-11
Changes to be compliant with the schema v0.5

alfredomp:
Changes to be compliant with the boutiques schema version 0.5.

We were using this example and realized that it wasn't compliant with the last version of the boutiques schema. We then made the changes so that future people testing boutiques with the github examples don't have errors when validating them.


gkiar:
Hey @alfredomp - thanks so much! Great & helpful contribution :)

gkiar:
(note to self and @glatard - while normally we merge into `dev`, since this is documentation and not touching any running code, I think it's important we merge it to master for better visibility of the change rather than wait until next release)

PR #282, opened 2018-07-30
Fix --changeUser (bosh -u)

simon-dube:
In its current version,  bosh was running useradd on the container to add the current user in case the option was specified. However, this would crash on certain container images that did not have the useradd command. Instead, we are now using the Docker -u option.


gkiar:
Hey @simon-dube looks great! Thanks!

Could you please sanitize the two lines that were too long and caused the build to fail, and then we'll merge?

Thanks!

gkiar:
Looks great - Thanks so much @simon-dube ! 😄

PR #290, opened 2018-08-09
Docker image inspection in creator

glatard:
Creator can now inspect a Docker image to fill some properties. In particular:
* If there is an entrypoint, it will be added to the command line
* Tool version is set from the image tag
* Author and description are set from the corresponding properties of the image

It also sets the container-image properties correctly, including when Docker images have to be booted through Singularity (`--use-singularity` option).

### DIFF
--- a/tools/python/boutiques/creator.py
+++ b/tools/python/boutiques/creator.py
@@ -21,10 +21,13 @@ class CreateDescriptor(object):
         with open(template) as f:
             self.descriptor = simplejson.load(f)
 
+        if(kwargs.get('docker_image')):
+            self.parse_docker(self.descriptor,
+                              kwargs.get('docker_image'),
+                              kwargs.get('use_singularity'))
+
         self.count = 0
-        if parser is None:
-            pass
-        else:
+        if parser is not None:
             self.parser = parser
             self.descriptor["inputs"] = []
             self.descriptor["tags"] = kwargs.get("tags") or {}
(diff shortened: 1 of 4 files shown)


tools/python/boutiques/creator.py
```diff
@@ -34,14 +40,64 @@ def __init__(self, parser=None, **kwargs):
...
+            cont_image['type'] = 'singularity'
+            cont_image['index'] = 'docker://'
+            cont_image['image'] = image_name
+        else:
+            cont_image['type'] = 'docker'
+            cont_image['image'] = image_name
+
+        # Properties found in the image metadata
+
+        ((stdout, stderr),
```
gkiar: Look into using singularity instead of Docker, if the user specified a singularity image.

gkiar: http://singularity.lbl.gov/docs-environment-metadata

glatard: Good idea! 
Note to self: here is how to find the entry point of a Singularity container:
```
$ singularity exec ./mysql.simg cat /.singularity.d/runscript
#!/bin/sh

exec "docker-entrypoint.sh" "$@"
```

glatard: In a first step, let's try to find metadata using Docker, and exit gracefully if Docker is not installed (or fall back on Singularity).

glatard:
Hey @gkiar I thought I would finish this before it conflicts with the main repo... So, for now, only Docker images can be specified, and image metadata won't be read unless Docker is installed. I created #306 to support Singularity images. Do you agree to merge?

gkiar:
Yep, I agree. Thanks!

PR #291, opened 2018-08-20
Fixed the syntax of the bosh import call at the top of the readme

shots47s:


glatard:
Here you go, I leave for vacations and they start merging directly into master... Just kidding, thanks for the fix!

Issue #268, opened 2018-05-10
new output structure is printed with simulate


gkiar:
When simulating, I get the below. We should make it so that this object is not returned to stdout when simulating.
```
$ bosh exec simulate ndmg_descriptor.json -r 2
Generated Commands:
mkdir -p ; ndmg_bids  ''   group --participant_label 'str_nn str_4o str_6F str_vd'  --atlas str_VN --minimal  --log
mkdir -p ; ndmg_bids  ''   participant --participant_label 'str_bX str_bI str_q4' --dataset str_Rr  --minimal  --log
Shell command

Container location

Container command

Exit code
0
Std out

Std err

Error message

Output files

Missing files

```

PR #301, opened 2018-09-11
Bug fix in simulate

gkiar:
SImulate now:
- supports descriptors with no outputs specified
- adds the field ID to the randomized value of string- and file-type inputs
- has another test to reflect the bullet above
- no longer prints the executor output log, since nothing was run and the status is always zero if it has gotten to that point


gkiar:
@glatard - ready for review; not sure why textcov went down since I didn't touch most of those lines, but it passes.

glatard:
Could you just refer to the issues it solves if any, and close them?

gkiar:
Also closes #268

Issue #281, opened 2018-07-30
add runtime properties to container-image


gkiar:
Something like the following two examples would allow users to specify container-level arguments for their application. The first is required for reprozip to be run inside Docker, and the second is for Nvidia drivers to be automagically mounted by Singularity.

These don't necessarily make sense as arguments to the descriptor, as they're fixed regardless of execution, but need to be captured somewhere so that there aren't "hidden" required dependencies for a bosh execution such as these specific mounts or flags passed on the command-line.

```
"container-image": {
  "image":"gkiar/test",
  "type": "docker",
  "index": "docker.io",
  "container-opts": ["--privileged", "--security-opt=seccomp:unconfined"]
}
...
"container-image": {
  "image":"gkiar/test",
  "type": "singularity",
  "index": "docker://",
  "container-opts": ["--nv"]
}
```

Issue #304, opened 2018-09-12
inputs can only belong to one group


leej3:
need to be able to belong to more

gkiar:
@glatard  an example would be a tool that has two inputs which must be paired (`all-or-none`), but are mutually exclusive with another options (`mutually-exclusive`).

I'd like to specify this as follows, assuming ids `2` and `3` must be used together, and are mutually exclusive with id `1`.

```
"groups": [
  {
    "id": "group1",
    "name": "group 1 name",
    "members": ["id1", "id2"],
    "mutually-exclusive": true
  },
  {
    "id": "group2",
    "name": "group 2 name",
    "members": ["id2", "id3"],
    "all-or-none": true
  }
]
```

For web forms, if they store the invocation based on input ID behind the scenes in JS, every field of an input could fetch the value from that object from memory, if it exists, and populate it. This just means some inputs could show up in two places, separated by group, but their values would be linked. Thoughts, @glatard ?

gkiar:
The above for displaying is not unlike what you suggested - I'm not worried about the complexity yet, since I haven't seen complex examples of this behaviour being necessary.

glatard:
Makes a lot of sense! Shall we then just stop checking that an input belongs to a single group in the validator?

gkiar:
I think we'll need to replace it with logic ensuring an input doesn't belong to conflicting groups, but some of that should already be baked in elsewhere and can be adopted (i.e. no mutually exclusive groups with required inputs)

erinb90:
So I thought of a couple rules we could implement once we start allowing inputs in multiple groups:

* Pairs of mutually exclusive inputs cannot both belong to an all-or-none group, and vice-versa, for example the following is invalid:
```
"groups": [
  {
    "id": "group1",
    "name": "group 1 name",
    "members": ["id2", "id3"],
    "mutually-exclusive": true
  },
  {
    "id": "group2",
    "name": "group 2 name",
    "members": ["id1", "id2", "id3"],
    "all-or-none": true
  }
]
```
* Pairs of inputs in an all-or-none group can't be the only inputs in a one-is-required group, for example the following is invalid:
```
"groups": [
  {
    "id": "group1",
    "name": "group 1 name",
    "members": ["id1", "id2", "id3"],
    "all-or-none": true
  },
  {
    "id": "group2",
    "name": "group 2 name",
    "members": ["id1", "id2"],
    "one-is-required": true
  }
]
```
[...]

Issue #277, opened 2018-07-16
Add support for JSON objects passed on the CLI in bosh


glatard:
In a curl way, e.g., `bosh exec launch descriptor.json '{"image": "/path/image.nii.gz", "param": 1}'`

glatard:
#213 implements that for `bosh exec launch`. The following functions of the CLI/Python API should also be modified:
- [ ] exec
- [ ] simulate
- [ ] import
- [ ] export
- [ ] publish
- [ ] invocation
- [ ] evaluate
- [ ] test

glatard:
@erinb90, was this solved by #323?

erinb90:
@glatard I think so, but I plan to write test cases for everything to confirm

Issue #263, opened 2018-05-02
Add an option to print stdout and stderr on the fly during execution


glatard:

gkiar:
such as `--stream`/`-s` in `bosh exec launch`, that in `localExec.py`, the `_localExecute` function, loops over `stdout` of the Popen process, like shown [here](https://stackoverflow.com/a/803421/1003437)

PR #309, opened 2018-09-14
#281 add runtime properties to container-image

erinb90:
Note: this branch has 5 extra commits that are on master but not on develop. Should develop be updated to have those?


tools/python/boutiques/schema/examples/example1/exampleTool1.py
```diff
@@ -40,12 +40,17 @@ def main(args=None):
...
         sys.exit(1)
-    
+   
     var = os.getenv("ENVAR")
     if var != "theValue":
         sys.stderr.write("error: ENVAR environment variable must be set to 'theValue'.\n")
         sys.exit(1)
 
+    hostname = os.getenv("HOSTNAME")
+    if hostname != "testHost":
```
erinb90: My issue right now is that this error gets thrown during the Singularity test since ```example1_sing.json``` doesn't set the hostname (which doesn't seem to be possible with Singularity). Is there a way to check in ```exampleTool1.py1``` whether it's being used by Docker or Singularity?

glatard: I think a Singularity image must contain `/.singularity.d/runscript` but @gkiar would know this better.

The additional 5 commits are typos that we agreed to push directly to master... but we probably shouldn't have and we won't do it again! 

Thanks for the PR btw, it looks good!

tools/python/boutiques/localExec.py
```diff
@@ -292,6 +293,11 @@ def execute(self, mount_strings):
             if launchDir is None:
                 launchDir = op.realpath('./')
             launchDir = op.realpath(launchDir)
+            # Get the container options
+            conOptsString = ""
+            if conOpts:
+                for opt in conOpts:
+                    conOptsString += opt + ' '
```
gkiar: In a perfect world, we could add some validation on what strings were passed, and that they're valid. But this requires options to be well documented by tools.

tools/python/boutiques/schema/examples/example1/example1_docker.json
```diff
@@ -3,7 +3,8 @@
     "command-line": "exampleTool1.py [CONFIG_FILE] [STRING_INPUT_LIST] [STRING_INPUT] [FLAG_INPUT] [NUMBER_INPUT] [ENUM_INPUT] [FILE_INPUT] [LOG] [INT_LIST_INPUT]",
     "container-image": {
         "image": "boutiques/example1:test",
-        "type": "docker"
+        "type": "docker",
+        "container-opts": ["--hostname testHost"]
```
gkiar: One convention, is that spaces are omitted and items follow one another in a list, i.e. `['--hostname', 'testHost']`. I'm not sure how we feel about this/care to copy it.

tools/python/boutiques/schema/examples/example1/exampleTool1.py
gkiar: Hey @glatard! @erinb90 and I just poked at this together, and will settle for setting & checking the `-e HOME=/tmp/somemadeuppath/`/`--home /tmp/somemadeuppath/` param in Docker/Singularity as a way to test this feature. We'll have to create the directory first, though, in the test. Thoughts?

glatard: Sounds great, the directory could be created in the Dockerfile used by example1.py

tools/python/boutiques/schema/examples/example1/example1_docker.json
glatard: I think we can just join the elements in container-opts with a space.

gkiar:
This looks awesome!

As discussed, some other potential issues to tackle, either in this PR or the next :) 

- https://github.com/boutiques/boutiques/issues/277 (Already in progress)
- https://github.com/boutiques/boutiques/issues/304
- https://github.com/boutiques/boutiques/issues/263

tools/python/boutiques/schema/examples/example1/exampleTool1.py
erinb90: I created a directory `/tmp/testDir/` in the Dockerfile, went in the container and saw it exists. However, when the Singularity test is run with the `--home /tmp/testDir/` option, I get the error `Home directory is not owned by calling user`. So I tried `--home $HOME:/tmp/testDir/` to map the source home to the destination home (see my latest changes), but am now getting a weird error:
```
WARNING: Could not chdir to home: /tmp/testDir/
/.exec: line 3: /home/travis/build/boutiques/boutiques/tools/python/temp-566401995773-1537479246275.localExec.boshjob.sh: No such file or directory
```
The temp file is in the run command, why would it not exist?

PR #310, opened 2018-09-21
#277 Add support for JSON objects passed on the CLI in bosh

erinb90:


tools/python/boutiques/validator.py
```diff
@@ -26,10 +26,17 @@ def validate_descriptor(json_file, **kwargs):
         schema = simplejson.load(fhandle)
 
     # Load descriptor
-    with open(json_file) as fhandle:
+    json_object = kwargs.get('json_object')
```
glatard: Check if it's a file (calling the dedicated function) instead of using an argument

tools/python/boutiques/validator.py
```diff
@@ -26,10 +26,17 @@ def validate_descriptor(json_file, **kwargs):
...
+            try:
+                descriptor = simplejson.load(fhandle)
+            except simplejson.errors.JSONDecodeError as e:
+                raise DescriptorValidationError(str(e))
+    else:
         try:
-            descriptor = simplejson.load(fhandle)
-        except simplejson.errors.JSONDecodeError as e:
+            descriptor = json.loads(json_file)
+        except ValueError as e:
```
glatard: Create a function that will return the JSON object or the JSON (on demand) from a string representing a JSON file or object. Use this function as much as possible.

glatard:
I'm merging this since it works, adding to #277 for the remaining methods.
Thanks for the good work @erinb90!

PR #313, opened 2018-09-27
Add an option to print stdout and stderr on the fly during execution

erinb90:
Issue #263 

Adding the `-s` or `--stream` option to `bosh exec launch` causes the executor to print stdout and stderr in real time


gkiar:
@erinb90 this looks awesome, thanks so much! @glatard - thoughts?

glatard:
This looks awesome indeed! 

I have tried it, but when I do:
```
$ bosh exec launch -s ./boutiques/schema/examples/good.json ./boutiques/schema/examples/good_invocation.json 
```
The stdout of `docker pull` doesn't get printed while the container image gets pulled. Any idea why?

tools/python/boutiques/localExec.py
```diff
@@ -435,7 +435,24 @@ def _localExecute(self, command):
             sys.stderr.write('Input Value Error during attempted execution!')
             raise e
         else:
```
glatard: Style: I don't think that `else` here is necessary, since the exceptions also raise exceptions. Removing `else` would save one level of indentation.

tools/python/boutiques/localExec.py
```diff
@@ -435,7 +435,24 @@ def _localExecute(self, command):
...
+                            sys.stdout.write(outLine + "\n")
+                            stdoutdata += outLine
+                        if errLine != '':
+                            sys.stderr.write(errLine + "\n")
+                            stderrdata += errLine
+                    else:
+                        break
+                return (stdoutdata, stderrdata), process.returncode
+            else:
+                return process.communicate(), process.returncode
```
glatard: Style nitpick: starting with `if not self.stream: return ...` would remove the need for `else` and would save one level of indentation. Likewise, in the `while`, doing `if not process.poll() is None: break` would remove the need for `else`. In general, try to avoid having more than 3 indentation levels (I know, some sections of this file are terrible in this respect).

tools/python/boutiques/tests/test_example1.py
```diff
@@ -71,6 +77,34 @@ def test_example1_exec_docker(self):
...
+        ret = bosh.execute("launch",
+                           os.path.join(example1_dir,
+                                        "example1_docker.json"),
+                           "-s",
+                           os.path.join(example1_dir,
+                                        "invocation.json"))
+
+        # Make sure stdout and stderr are printed on the fly for
+        # streaming mode
+        out, err = self.capfd.readouterr()
```
glatard: Sorry, I don't understand what this does :) Could you add a line to explain where readouterr() comes from?

erinb90:
> This looks awesome indeed!
> 
> I have tried it, but when I do:
> 
> ```
> $ bosh exec launch -s ./boutiques/schema/examples/good.json ./boutiques/schema/examples/good_invocation.json 
> ```
> 
> The stdout of `docker pull` doesn't get printed while the container image gets pulled. Any idea why?

@glatard It does get printed on the fly when using `-s` but doesn't get printed in the final results at the end. This is because the return value of `self._localExecute("docker pull " + str(conImage))` (containing the stdout) isn't stored anywhere. Only the stdout of the container command is stored and printed. This is a separate issue having nothing to do with my changes.

tools/python/boutiques/tests/test_example1.py
erinb90: `capfd` is a Pytest fixture that captures the stdout and stderr during test execution and returns them as a tuple in `readouterr()`. I'll add another comment in the file explaining this.

https://docs.pytest.org/en/latest/capture.html

glatard:
I tried again, now I have this:
```
$ bosh exec launch -s ./boutiques/schema/examples/good.json ./boutiques/schema/examples/good_invocation.json 
  File "/home/glatard/virtualenvs/boutiques/bin/bosh", line 11, in <module>
    sys.exit(bosh())
  File "/home/glatard/virtualenvs/boutiques/lib/python3.6/site-packages/boutiques/bosh.py", line 424, in bosh
    out = execute(*params)
  File "/home/glatard/virtualenvs/boutiques/lib/python3.6/site-packages/boutiques/bosh.py", line 125, in execute
    return executor.execute(results.volumes)
  File "/home/glatard/virtualenvs/boutiques/lib/python3.6/site-packages/boutiques/localExec.py", line 244, in execute
    if self._localExecute("docker pull " + str(conImage))[1]:
  File "/home/glatard/virtualenvs/boutiques/lib/python3.6/site-packages/boutiques/localExec.py", line 451, in _localExecute
    sys.stdout.write(outLine)
TypeError: write() argument must be str, not bytes
```
I'm not sure how the tests pass while this crashes, there must be a glitch somewhere.

I'm looking forward to using it, this will be very useful!

glatard:
any update on this? Now it's conflicting, we should aim at merging it soon.

erinb90:
@glatard Sorry I thought I had fixed this but I didn't (maybe I dreamed it). I think it was breaking because I had removed the thing that decodes the bytes into a string (because I saw the tests passed without it). Adding it back now :)

PR #315, opened 2018-10-03
Prevent endless growing of [] brackets in creator

gkiar:
If [] brackets are already there, don't add them again :)


glatard:
Looks great, thanks!

Issue #317, opened 2018-10-04
Can't have multiple tags with the same key


erinb90:
For example, if you want to publish a descriptor with tags:
```
"tags": {
    "domain": "neuroinformatics",
    "domain": "fmri"
}
```
Only the last one (`domain:fmri`) will be published to Zenodo. We should support having more than one tag with the same key.

Issue #311, opened 2018-09-24
bosh publish fails if there is no flag defined


camarasu:
Here's the error log:

$ bosh publish --sandbox --no-int -v noTagsApp.json
[ INFO ] Zenodo access token saved in /home/camarasu/.boutiques
[ INFO ] Using Zenodo endpoint https://sandbox.zenodo.org
[ INFO (401) ] Zenodo is accessible
[ INFO (200) ] Authentication to Zenodo successful
Traceback (most recent call last):
  File "/usr/bin/bosh", line 11, in <module>
    sys.exit(bosh())
  File "/home/camarasu/.local/lib/python2.7/site-packages/boutiques/bosh.py", line 420, in bosh
    out = publish(*params)
  File "/home/camarasu/.local/lib/python2.7/site-packages/boutiques/bosh.py", line 248, in publish
    publisher.publish()
  File "/home/camarasu/.local/lib/python2.7/site-packages/boutiques/publisher.py", line 204, in publish
    deposition_id = self.zenodo_deposit()
  File "/home/camarasu/.local/lib/python2.7/site-packages/boutiques/publisher.py", line 126, in zenodo_deposit
    for tag in self.descriptor.get('tags'):
TypeError: 'NoneType' object is not iterable

PR #318, opened 2018-10-05
Publisher fixes

erinb90:
* Can now publish multiple tags with the same key (#317)
    * Can specify tag values as either a string or a list of strings
    * Changed example1_docker tags to include "foo: ["bar1", "bar2"]" to test this functionality (see screenshot)
* App no longer crashes when trying to publish a descriptor with no tags (#311)
* App no longer crashes when you say no to the "Are you sure" message when publishing in interactive mode.
* The "Are you sure (Y/n)" when publishing is no longer case sensitive (can say "y" instead of "Y" - I think this makes more sense)
* Fixed a couple typos

zenodo tags


tools/python/boutiques/schema/descriptor.schema.json
```diff
@@ -654,7 +654,10 @@
             "type": "object",
             "description": "An set of key-value pairs specifying tags describing the pipeline. The tag names are open, they might be more constrained in the future.",
             "additionalProperties": {
-                "type": "string",
+                "type": ["string", "array"],
+                "items": {
+                    "type": "string"
+                },
```
glatard: This is not backward compatible, which we try to avoid whenever possible. I think it would make sense that a tag can be a string or an array, so that users can still specify just a string when they don't have multiple values. Could you update the schema and the tests accordingly?

erinb90: It already allows for both. `"type": ["string", "array"]` means that the value can be either a string or an array. Also, the `"items"` restriction will only be applied if the value is an array and will be ignored if it's a string. It's explained here: https://cswr.github.io/JsonSchema/spec/multiple_types/

glatard: Sorry I had missed that, that's perfect!

erinb90:
The Travis builds are failing at
```
$ mv ${HOME}/${IMGNAME}.img ./boutiques-example1-test.simg
mv: cannot stat ‘/home/travis/.img’: No such file or directory
```
Any idea why? I didn't touch that part of the code.

glatard:
It must be an update in the Docker 2 Singularity converter that led IMGNAME to be undefined.

erinb90:
> It must be an update in the Docker 2 Singularity converter that led IMGNAME to be undefined.
> 
> Perhaps you could run the docker2singularity conversion locally and adjust the definition of IMGNAME in the travis file accordingly. There was an update of docker2singularity recently. We should use version 2.6 and fix the Docker tag to 2.6 too.

Everything works fine locally, so I'm not quite sure what needs to be changed.

glatard:
The problem was in the `grepping` of `IMGNAME` in the output of `singularity2docker`. I fixed it in https://github.com/glatard/boutiques/tree/singularity2docker, could you pull this branch into yours (also in https://github.com/boutiques/boutiques/pull/321)? Only the `.travis.yml` changed. Now I assume that the Singularity image will be produced in $HOME, as passed in the `Docker` mount dirs, that's safer than the `grep+awk+cut+cut` done previously.

Issue #299, opened 2018-09-10
Version consolidation on Zenodo


glatard:
`bosh publish` should publish different versions of the same tool descriptor in a consolidated way on Zenodo. This seems to be possible, see for instance the "Versions" box in the Boutiques page on Zenodo: https://zenodo.org/record/1409707

Issue #241, opened 2018-04-17
bosh search


glatard:
`bosh search` will search for boutiques descriptors on Zenodo. Descriptors will be retrieved from the `boutiques` keyword and possibly other keywords such as `schema-version` or `container-image`. This should be implemented using the Records search API, currently in testing (see http://developers.zenodo.org/#rest-api).

Easy extensions to `search` could include `pull` and run from descriptor, following the Docker client model.

glatard:
`bosh search` should behave as `docker search`. We could use the Zenodo ID as key in the returned list. Then, keys should be usable in other commands, such as `bosh exec`, etc.

glatard:
The Zenodo query to get Boutiques descriptors should be:
q=boutiques&keywords=boutiques&keywords=schema&keywords=version&file_type=json&type=software

Hopefully the Records API allows for that. If not, let's hack something from the URL.

erinb90:
This issue can be closed.

PR #322, opened 2018-10-18
Version consolidation on Zenodo

erinb90:
Issue #299 

Can now publish an updated version of an existing descriptor using the flag `--update <descriptor_id>` on `bosh publish`, where `descriptor_id` is the Zenodo ID of the published record you wish to update.

Example:
zenodo versions

### DIFF
--- a/tools/python/boutiques/publisher.py
+++ b/tools/python/boutiques/publisher.py
@@ -239,9 +239,9 @@ class Publisher():
         }
         keywords = data['metadata']['keywords']
         if self.descriptor.get('tags'):
-            for key, value in self.descriptor.get('tags').iteritems():
+            for key, value in self.descriptor.get('tags').items():
                 # Check if value is a string or a list of strings
-                if isinstance(value, basestring):
+                if isinstance(value, str):
                     keywords.append(key + ":" + value)
                 else:
                     keywords += [key + ":" + item for item in value]
--- a/tools/python/boutiques/publisher.py
+++ b/tools/python/boutiques/publisher.py
@@ -109,51 +109,7 @@ class Publisher():
 
     def zenodo_deposit(self):
         headers = {"Content-Type": "application/json"}
-        data = {
-            'metadata': {
-                'title': self.descriptor['name'],
-                'upload_type': 'software',
-                'description': self.descriptor['description'] or "Boutiques "
-                               "descriptor for {0}".format(
-                                                   self.descriptor['name']),
-                'creators': [{'name': self.creator}],
-                'version': self.descriptor['tool-version'],
-                'keywords': ['Boutiques',
-                             'schema-version:{0}'.
-                             format(self.descriptor['schema-version'])]
-            }
-        }
-        keywords = data['metadata']['keywords']
-        if self.descriptor.get('tags'):
-            for key, value in self.descriptor.get('tags').items():
-                # Check if value is a string or a list of strings
-                if isinstance(value, str):
-                    keywords.append(key + ":" + value)
-                else:
-                    keywords += [key + ":" + item for item in value]
-        if self.descriptor.get('container-image'):
-            keywords.append(self.descriptor['container-image']['type'])
-        if self.url is not None:
-            if data['metadata'].get('related_identifiers') is None:
-                data['metadata']['related_identifiers'] = []
-            data['metadata']['related_identifiers'].append({
-                'identifier': self.url,
-                'relation': 'hasPart'
-            })
-        if self.tool_doi is not None:
-            if data['metadata'].get('related_identifiers') is None:
-                data['metadata']['related_identifiers'] = []
-            data['metadata']['related_identifiers'].append({
-                'identifier': self.tool_doi,
-                'relation': 'hasPart'
-            })
-        if self.descriptor_url is not None:
-            if data['metadata'].get('related_identifiers') is None:
-                data['metadata']['related_identifiers'] = []
-            data['metadata']['related_identifiers'].append({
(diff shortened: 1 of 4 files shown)


glatard:
hi @erinb90, this is superb! A few comments at this stage:
1. We need to be able to 'detect' that a descriptor is an update of another one. Otherwise this will still be error-prone and people will end-up with redundant descriptors anyway. Perhaps you could search for a boutiques tool with the same tool name. The client would then ask confirmation before updating it (and assume 'yes' in non-interactive mode).
2. What happens if we update with the same tool version? Do we see 2 identical versions in the box on the right? It is fine if it behaves like this, just checking.
3. 1 requires a basic version of `search` (#241). I guess `search` is your next step!

erinb90:
> Hi @erinb90, this is superb! A few comments:
> 
> 1. We need to be able to 'detect' that a descriptor is an update of another one. Otherwise this will still be error-prone and people will end-up with redundant descriptors anyway. Perhaps you could search for a boutiques tool with the same tool name. The client would then ask confirmation before updating it (and assume 'yes' in non-interactive mode).
> 2. What happens if we update with the same tool version? Do we see 2 identical versions in the box on the right? It is fine if it behaves like this, just checking.
> 3. 1 requires a basic version of `search` (#241). I guess `search` is your next step!

@glatard 
1. Yes, I remember you talking last week about finding the descriptor to update using a search. I will update this PR to search Zenodo for a descriptor with a matching title to the one passed to `bosh publish` and, if it finds one, use its ID to publish an updated version. 
2. Yes it will show 2 identical versions in the box. The updated one uses whatever we put as the tool version even if it exists already.
3. #241 will be completed at the same time as I update this PR.

tools/python/boutiques/tests/test_publisher.py
```diff
@@ -52,6 +62,45 @@ def test_publication(self):
...
+        with open(temp_descriptor_updated.name, 'r') as fhandle:
+            descriptor_updated = json.load(fhandle)
+            assert(descriptor_updated.get('doi') == new_doi)
+            assert(descriptor_updated.get('doi') != doi)
+
+        # Search for the descriptor by its title
+        results = bosh(["search", unique_title, "--sandbox", "-v"])
+
+        # Make sure there is only one result with this title
+        # and that its DOI is the updated one
```
erinb90: So this seems to randomly fail on Travis... sometimes the search doesn't find the descriptor even though in the logs you can see that it was published successfully. Any ideas why?

glatard: not really, most likely their API is a bit unstable. This is related to #289. @simon-dube told me that good solutions exist to mock servers but I don't remember which ones...

erinb90: Ok, so should I remove the test?

simon-dube: Hi @glatard, sorry for the delay. I don't remember using mocks in the CARMIN server project, but a unit test should not depend on remote calls. As a side note, from what I can see, I believe that testing that publishing a descriptor which alrady has a DOI should fail should be moved to its own test.

tools/python/boutiques/publisher.py
```diff
@@ -169,6 +125,21 @@ def zenodo_deposit(self):
...
+                          % deposition_id,
+                          params={'access_token': self.zenodo_access_token})
+        if(r.status_code != 201):
+            self.raise_zenodo_error("Deposition of new version failed", r)
+        if(self.verbose):
+            self.print_zenodo_info("Deposition of new version succeeded", r)
+        new_url = r.json()['links']['latest_draft']
+        new_zid = new_url.split("/")[-1]
+        self.zenodo_update_metadata(new_zid)
+        self.zenodo_delete_files(new_zid, r.json()["files"])
```
glatard: what are you deleting here?

tools/python/boutiques/publisher.py
```diff
@@ -215,9 +212,105 @@ def publish(self):
...
+                prompt = ("Found an existing record with the same name, "
+                          "would you like to update it? "
+                          "(Y:update existing/n:publish new version)")
+                try:
+                    ret = raw_input(prompt)  # Python 2
+                except NameError:
+                    ret = input(prompt)  # Python 3
+                if ret.upper() != "Y":
+                    self.publish_new_entry()
+            self.publish_updated_version(id_to_update)
```
glatard: I think you're missing an `else` statement here: I tried to "publish a new version" and it created a new entry and a new version.

tools/python/boutiques/publisher.py
```diff
@@ -215,9 +212,105 @@ def publish(self):
...
+            title = hit["metadata"]["title"]
+            if title == self.descriptor.get("name"):
+                id_to_update = hit["id"]
+                break
+
+        if id_to_update:
+            if(not self.no_int):
+                prompt = ("Found an existing record with the same name, "
+                          "would you like to update it? "
+                          "(Y:update existing/n:publish new version)")
```
glatard: I would say `"Publish new entry with name {}".format(name)` to make it clearer. Also, add a space after the closing bracket.

tools/python/boutiques/publisher.py
erinb90: When you create a new version, Zenodo copies everything from the previous version, including the uploaded files. So here I am deleting any files from the previous version before uploading the updated file. If you don't delete the files and you try to upload one with the same name, their server throws a 400 error. If you don't delete and you try to upload one with a different name, the new version will have both the old and the new version, which I would assume the user does not want.

PR #323, opened 2018-10-26
Implemented bosh search and bosh pull, added ability to run directly from zenodo

erinb90:
Issue #241 

`bosh search` searches Zenodo for all descriptors and returns a list like this:
image

`bosh search <search_query>` searches for a specific descriptor, for example `bosh search ndmg` will return:
image

`bosh pull <ZID>` downloads a descriptor from Zenodo using its ID prefixed with 'zenodo.' (as returned from bosh search), e.g. `bosh pull zenodo.1472823`

`bosh exec launch <ZID> <invocation>` launches an invocation using the pull functionality to open and use the descriptor directly from its published entry on Zenodo.

### DIFF
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -376,6 +376,22 @@ def search(*params):
     return searcher.search()
 
 
+def pull(*params):
+    parser = ArgumentParser("Download a descriptor from Zenodo.")
+
+    parser.add_argument("zid", action="store", help="Zenodo ID")
+
+    parser.add_argument("-v", "--verbose", action="store_true",
+                        help="Print information messages")
+
+    result = parser.parse_args(params)
+
+    from boutiques.puller import Puller
+    puller = Puller(result.zid, result.verbose)
+
+    return puller.pull()
+
+
 def bosh(args=None):
     parser = ArgumentParser(description="Driver for Bosh functions",
                             add_help=False)
@@ -394,10 +410,11 @@ def bosh(args=None):
                         "given descriptor. Eval: given an invocation and a "
                         "descriptor, queries execution properties. "
                         "Test: run pytest on a descriptor detailing tests. "
-                        "Search: search Zenodo for descriptors.",
+                        "Search: search Zenodo for descriptors. "
+                        "Pull: download a descriptor from Zenodo.",
                         choices=["create", "validate", "exec", "import",
                                  "export", "publish", "invocation", "evaluate",
-                                 "test", "search"])
+                                 "test", "search", "pull"])
 
     parser.add_argument("--help", "-h", action="store_true",
                         help="show this help message and exit")
@@ -457,6 +474,9 @@ def bosh(args=None):
             return bosh_return(out)
         elif func == "search":
             out = search(*params)
+            return bosh_return(out)
+        elif func == "pull":
+            out = pull(*params)
             return bosh_return(out)
         else:
             parser.print_help()
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -99,16 +99,12 @@ def execute(*params):
         parser.add_argument("-s", "--stream", action="store_true",
                             help="Streams stdout and stderr in real time "
                             "during execution.")
-        parser.add_argument("-z", "--zenodo", action="store_true",
-                            help="Download and use a descriptor published "
-                            "on Zenodo.")
         results = parser.parse_args(params)
         descriptor = results.descriptor
         inp = results.invocation
 
         # Validate invocation and descriptor
-        if not results.zenodo:
-            valid = invocation(descriptor, '-i', inp)
+        valid = invocation(descriptor, '-i', inp)
 
         # Generate object that will perform the commands
         from boutiques.localExec import LocalExecutor
@@ -116,8 +112,7 @@ def execute(*params):
                                  {"forcePathType": True,
                                   "debug": results.debug,
                                   "changeUser": results.user,
-                                  "stream": results.stream,
-                                  "zenodo": results.zenodo})
+                                  "stream": results.stream})
         # Execute it
         return executor.execute(results.volumes)
 
@@ -384,8 +379,9 @@ def search(*params):
 def pull(*params):
     parser = ArgumentParser("Download a descriptor from Zenodo.")
 
-    parser.add_argument("identifier", action="store", help="Zenodo ID or "
-                        "filename of the descriptor to pull")
+    parser.add_argument("identifier", action="store", help="Zenodo ID "
+                        "of the descriptor to pull, prefixed by "
+                        "'zenodo', e.g. zenodo.123456")
 
     parser.add_argument("-v", "--verbose", action="store_true",
                         help="Print information messages")
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -395,7 +395,7 @@ def pull(*params):
     from boutiques.puller import Puller
     puller = Puller(result.identifier, result.verbose, True)
 
-    return puller.pull()
+    puller.pull()
 
 
 def bosh(args=None):
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -396,7 +396,7 @@ def pull(*params):
     result = parser.parse_args(params)
 
     from boutiques.puller import Puller
-    puller = Puller(result.identifier, result.verbose, True)
+    puller = Puller(result.zid, result.verbose, True)
 
     puller.pull()
 
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -144,9 +144,6 @@ def execute(*params):
             raise SystemExit("Input file {} does not exist.".format(inp))
         if inp and not inp.endswith(".json"):
             raise SystemExit("Input file {} must end in 'json'.".format(inp))
-        # if not os.path.isfile(descriptor):
-        #     raise SystemExit("JSON descriptor {} does not seem to exist."
-        #                      .format(descriptor))
         if not rand and not inp:
             raise SystemExit("The default mode requires an input (-i).")
 
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -358,27 +358,46 @@ def test(*params):
     return pytest.main([test_path, "--descriptor", result.descriptor])
 
 
+def search(*params):
+    parser = ArgumentParser("Search Zenodo for Boutiques descriptors. "
+                            "When no term is supplied, will search for "
+                            "all descriptors.")
+
+    parser.add_argument("-q", "--query", action="store", help="Search query")
+
+    parser.add_argument("-v", "--verbose", action="store_true",
+                        help="Print information messages")
+
+    result = parser.parse_args(params)
+
+    from boutiques.searcher import Searcher
+    searcher = Searcher(result.query, result.verbose)
+
+    return searcher.search()
+
+
 def bosh(args=None):
     parser = ArgumentParser(description="Driver for Bosh functions",
                             add_help=False)
     parser.add_argument("function", action="store", nargs="?",
                         help="The tool within boutiques/bosh you wish to run. "
-                        "Create: creates an Boutiques descriptor from scratch."
-                        "Validate: validates an existing boutiques descriptor."
+                        "Create: creates an Boutiques descriptor from scratch. "
+                        "Validate: validates an existing boutiques descriptor. "
                         "Exec: launches or simulates an execution given a "
                         "descriptor and a set of inputs. Import: creates a "
                         "descriptor for a BIDS app or updates a descriptor "
                         "from an older version of the schema. Export: exports a"
                         "descriptor to other formats. Publish: creates"
                         "an entry in Zenodo for the descriptor and "
-                        "adds the DOI created by Zenodo to the descriptor."
+                        "adds the DOI created by Zenodo to the descriptor. "
                         "Invocation: generates the invocation schema for a "
                         "given descriptor. Eval: given an invocation and a "
-                        "descriptor, queries execution properties."
-                        "Test: run pytest on a descriptor detailing tests",
+                        "descriptor, queries execution properties. "
+                        "Test: run pytest on a descriptor detailing tests. "
+                        "Search: search Zenodo for descriptors.",
                         choices=["create", "validate", "exec", "import",
                                  "export", "publish", "invocation", "evaluate",
-                                 "test"])
+                                 "test", "search"])
 
     parser.add_argument("--help", "-h", action="store_true",
                         help="show this help message and exit")
@@ -435,6 +454,9 @@ def bosh(args=None):
             return bosh_return(out)
         elif func == "test":
             out = test(*params)
+            return bosh_return(out)
+        elif func == "search":
+            out = search(*params)
             return bosh_return(out)
         else:
             parser.print_help()
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -17,6 +17,7 @@ from boutiques.localExec import ExecutorError
 from boutiques.exporter import ExportError
 from boutiques.importer import ImportError
 from boutiques.localExec import loadJson
+from tabulate import tabulate
 
 
 def create(*params):
@@ -395,7 +396,7 @@ def pull(*params):
     from boutiques.puller import Puller
     puller = Puller(result.zid, result.verbose, True)
 
-    puller.pull()
+    return puller.pull()
 
 
 def bosh(args=None):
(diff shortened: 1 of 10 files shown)


tools/python/boutiques/bosh.py
```diff
@@ -99,20 +99,25 @@ def execute(*params):
...
+        parser.add_argument("-z", "--zenodo", action="store_true",
+                            help="Download and use a descriptor published "
+                            "on Zenodo.")
         results = parser.parse_args(params)
         descriptor = results.descriptor
         inp = results.invocation
 
         # Validate invocation and descriptor
-        valid = invocation(descriptor, '-i', inp)
+        if not results.zenodo:
```
erinb90: this should actually have a TODO on it. I need to allow `bosh validate` to use a descriptor from Zenodo as well.

glatard:
Hih @erinb90, I will make more detailed comments later but I already have one: this.is.awesome! :100: :+1: :1st_place_medal:

erinb90:
fyi I just closed and reopened the PR to trigger a new Travis build now that the correct descriptor is on Zenodo.

tools/python/boutiques/bosh.py
```diff
@@ -99,20 +99,25 @@ def execute(*params):
         parser.add_argument("-s", "--stream", action="store_true",
                             help="Streams stdout and stderr in real time "
                             "during execution.")
+        parser.add_argument("-z", "--zenodo", action="store_true",
```
glatard: Ideally we wouldn't like to have to specify '-z' here, just like Docker can run from an image ID or name. Perhaps we could just prefix Zenodo IDs with `zenodo.`, as done in DOIs minted by Zenodo, to make it clear. In this case, `bosh search` should prefix the IDs with `zenodo.` so that the user could just copy-paste them on their command lines. Then if there is a file starting with `zenodo.`, then too bad, we'll ignore it. Files have to end with `.json` anyway, which won't be the case of a Zenodo id.

tools/python/boutiques/bosh.py
glatard: Pretty much all the functions have to accept Zenodo IDs then. This behavior could be factorized in a function that would accept a file, a JSON string, or a Zenodo id, and return a JSON object or dictionary representing the descriptor. All the methods would call this function to parse the descriptor or invocation. Then we could easily add more representations or behavior in the future, for instance caches, etc.

tools/python/boutiques/bosh.py
```diff
@@ -358,27 +363,64 @@ def test(*params):
     return pytest.main([test_path, "--descriptor", result.descriptor])
 
 
+def search(*params):
+    parser = ArgumentParser("Search Zenodo for Boutiques descriptors. "
+                            "When no term is supplied, will search for "
+                            "all descriptors.")
+
+    parser.add_argument("-q", "--query", action="store", help="Search query")
```
glatard: I would make it a positional argument.

tools/python/boutiques/localExec.py
```diff
@@ -132,8 +132,21 @@ def __init__(self, desc, invocation, options={}):
...
+        # Include: forcePathType and debug
+        self.debug = False
+        self.zenodo = False
+        for option in list(options.keys()):
+            setattr(self, option, options.get(option))
+
+        if self.zenodo:
+            from boutiques.puller import Puller
+            puller = Puller(desc, self.debug, False)
+            self.desc_dict = json.loads(puller.pull().read())
```
glatard: We should be careful with reproducibility here. We need to make sure that users won't be running a different analysis if a new descriptor is pushed between two runs. My feeling is that Zenodo ids are immutable, i.e., publishing a new version of the descriptor would result in a new id. Is it indeed the case? 

Note: Docker container images are not immutable, one might still mess up reproducibility that way, but we shouldn't add to this issue here.

tools/python/boutiques/localExec.py
```diff
@@ -132,8 +132,21 @@ def __init__(self, desc, invocation, options={}):
...
+        else:
+            # Parse JSON descriptor
+            self.desc_dict = loadJson(desc)
```
glatard: The logic above could be factorized in a function called by whoever needs to access a descriptor, as mentioned earlier.

tools/python/boutiques/puller.py
```diff
@@ -0,0 +1,77 @@
...
+    def pull_by_id(self, r, zid):
+        for hit in r.json()["hits"]["hits"]:
+            file_path = hit["files"][0]["links"]["self"]
+            file_name = file_path.split("/")[-1]
+            if hit["id"] == int(zid):
+                if self.download:
+                    if(self.verbose):
+                        self.print_zenodo_info("Downloading descriptor %s"
+                                               % file_name, r)
+                    return urlretrieve(file_path, file_name)
```
glatard: It would be nice to store the descriptors in some common place, for instance under `.boutiques` (but be careful, `.boutiques` is currently a file containing Zenodo tokens), so that they can be cached. If descriptors are immutable, caching won't create additional issues. Caching would also allow to work offline seamlessly or be robust to Zenodo downtimes. It should be easy enough to do, but feel free to add this as an issue for future work.

tools/python/boutiques/puller.py
```diff
@@ -0,0 +1,77 @@
...
+                if(self.verbose):
+                    self.print_zenodo_info("Opening descriptor %s"
+                                           % file_name, r)
+                return urlopen(file_path)
+
+        raise IOError("Descriptor not found")
+
+    def print_zenodo_info(self, message, r):
+        print("[ INFO ({1}) ] {0}".format(message, r.status_code))
```
glatard: We really should have a common logger throughout the project. Could you add an issue on that?

tools/python/boutiques/searcher.py
```diff
@@ -0,0 +1,46 @@
...
+                         'keywords=boutiques&keywords=schema&'
+                         'keywords=version&file_type=json&type=software'
+                         % self.query)
+        if(r.status_code != 200):
+            self.raise_zenodo_error("Error searching Zenodo", r)
+
+        if(self.verbose):
+            self.print_zenodo_info("Zenodo search returned %d results"
+                                   % r.json()["hits"]["total"], r)
+
```
glatard: Code seems redundant with Puller, perhaps factorize it?

tools/python/boutiques/searcher.py
```diff
@@ -0,0 +1,46 @@
...
+        return self.create_results_table(r.json())
+
+    def create_results_table(self, results):
+        table = []
+        for hit in results["hits"]["hits"]:
+            table.append([hit["id"], hit["metadata"]["title"],
+                         hit["metadata"]["description"]])
+        return tabulate(table, headers=['ID', 'TITLE',
+                        'DESCRIPTION'], tablefmt='plain')
```
glatard: A verbose mode could return more info, such as:
* Tags
* Author
* Container ('Docker', 'Singularity' or 'None')
* Schema version
* DOI
* Tests y/n
* Tool version
* Descriptor version
Some of these things should be accessible through Zenodo without having to download the descriptor. For the other ones, we could add explicit Zenodo tags in the publisher (create an issue).

glatard: Also, verbose could list all the versions of the tools, while non-verbose could aggregate them, as it is the case now.

tools/python/boutiques/bosh.py
erinb90: I already have the function that checks if the descriptor is a JSON file or object, so I can just extend that further to check if it's a Zenodo ID.

tools/python/boutiques/bosh.py
erinb90: Ok, but that would make it mandatory, right? I made it optional so that you could just type `bosh search` and see all the descriptors.

tools/python/boutiques/localExec.py
erinb90: Yes, a new Zenodo ID is generated every time you publish a new version.

tools/python/boutiques/bosh.py
glatard: nope, there can be optional positional arguments: https://stackoverflow.com/questions/4480075/argparse-optional-positional-arguments

tools/python/boutiques/searcher.py
glatard: Actually non-verbose should report only the latest versions.

tools/python/boutiques/puller.py
erinb90: I looked into this and it seems like it will cause problems with the publisher, which loads the entire `.boutiques` file to read the access tokens. Rather than refactoring the publisher to accommodate this, what if we cached the descriptors in the user's `.cache` folder?

glatard: You are right, using `.cache` makes much more sense! Note that `os.getenv("HOME")` wouldn't work on Windows (which we don't support yet but might in the future). Here is how to get the HOME cross-platform: https://stackoverflow.com/questions/2668909/how-to-find-the-real-user-home-directory-using-python If you're doing it, could you also update the publisher to not use `os.gentenv("HOME")`?

tools/python/boutiques/puller.py
```diff
@@ -0,0 +1,61 @@
...
+        from boutiques.searcher import Searcher
+        searcher = Searcher(None, self.verbose)
+        r = searcher.zenodo_search()
+
+        for hit in r.json()["hits"]["hits"]:
+            file_path = hit["files"][0]["links"]["self"]
+            file_name = file_path.split("/")[-1]
+            if hit["id"] == int(self.zid):
+                if self.download:
+                    cache_dir = os.path.join(os.getenv("HOME"), ".cache",
```
glatard: see my earlier comment on `getenv("HOME")`.

tools/python/boutiques/bosh.py
```diff
@@ -141,9 +144,9 @@ def execute(*params):
             raise SystemExit("Input file {} does not exist.".format(inp))
         if inp and not inp.endswith(".json"):
             raise SystemExit("Input file {} must end in 'json'.".format(inp))
-        if not os.path.isfile(descriptor):
-            raise SystemExit("JSON descriptor {} does not seem to exist."
-                             .format(descriptor))
+        # if not os.path.isfile(descriptor):
+        #     raise SystemExit("JSON descriptor {} does not seem to exist."
+        #                      .format(descriptor))
```
glatard: you might as well just remove these lines

tools/python/boutiques/bosh.py
```diff
@@ -436,6 +481,12 @@ def bosh_return(val, code=0, hide=False):
         elif func == "test":
             out = test(*params)
             return bosh_return(out)
+        elif func == "search":
+            out = search(*params)
+            return bosh_return(out)
```
glatard: so in case `search` is called from the Python API, it would return a tabulated string, difficult to parse. It would be better to return a dict in this case. Suggestion:
1. search could return a dict
2. bosh_return could take a formatting function as a parameter, defaulting to `print`, and set to `print(tabulate(x))` in this case.

tools/python/boutiques/bosh.py
```diff
@@ -358,27 +364,66 @@ def test(*params):
...
+
+    parser.add_argument("-v", "--verbose", action="store_true",
+                        help="Print information messages")
+
+    result = parser.parse_args(params)
+
+    from boutiques.puller import Puller
+    puller = Puller(result.zid, result.verbose, True)
+
+    puller.pull()
```
glatard: should be returned

tools/python/boutiques/bosh.py
```diff
@@ -436,6 +481,12 @@ def bosh_return(val, code=0, hide=False):
         elif func == "test":
             out = test(*params)
             return bosh_return(out)
+        elif func == "search":
+            out = search(*params)
+            return bosh_return(out)
+        elif func == "pull":
+            out = pull(*params)
+            return bosh_return(out)
```
glatard: not sure if we want to print the value returned by pull when used in CLI. If not, set `hide` to `True`.

tools/python/boutiques/localExec.py
```diff
@@ -1077,14 +1079,21 @@ def check(keyname, isGood, msg, value):  # Checks input values
...
+        from boutiques.puller import Puller
+        puller = Puller(userInput, False, False)
+        return json.loads(puller.pull().read())
+    # JSON object
     else:
         try:
-            return json.loads(jsonInput)
+            return json.loads(userInput)
         except ValueError:
             raise ExecutorError("Unable to decode JSON object")
```
glatard: That might be confusing: if the file path is wrong, error will be "Unable to decode JSON object". I would put a more detailed: `Cannot parse descriptor {}: file not found, invalid Zenodo ID, or invalid JSON object".format(userInput)`.

tools/python/boutiques/puller.py
```diff
@@ -0,0 +1,61 @@
...
+        self.download = download
+
+    def pull(self):
+        from boutiques.searcher import Searcher
+        searcher = Searcher(None, self.verbose)
+        r = searcher.zenodo_search()
+
+        for hit in r.json()["hits"]["hits"]:
+            file_path = hit["files"][0]["links"]["self"]
+            file_name = file_path.split("/")[-1]
```
glatard: use `os.sep` instead of `/`.

tools/python/boutiques/tests/test_pull.py
```diff
@@ -0,0 +1,22 @@
...
+from unittest import TestCase
+from boutiques.puller import ZenodoError
+import os
+
+
+class TestPull(TestCase):
+
+    def test_pull(self):
+        bosh(["pull", "zenodo.1472823"])
+        cache_dir = os.path.join(os.getenv("HOME"), ".cache", "boutiques")
```
glatard: see previous comment on `HOME`.

glatard:
See a few additional comments in the code. This is starting to look really good!

tools/python/boutiques/bosh.py
erinb90: oops, forgot to remove that

tools/python/boutiques/searcher.py
erinb90: Aren't tool version and descriptor version the same thing (at least with the way publisher is currently implemented)?

erinb90: Also, getting all versions of the tool isn't really possible through the search API. The JSON returned by the search only lists the current version's ID, the number of versions and the original version's ID. Should I include any of those?

erinb90: The tabular format looks a bit messy on my little VM with all those columns:
image
Should I keep it this way or change the format?

tools/python/boutiques/puller.py
erinb90: Since Windows doesn't have a `.cache` folder (cache is stored in user/AppData), I'm thinking maybe we could just create a `.boutiques` directory in the user's home dir that would store the config file with the Zenodo tokens as well as the pulled descriptors.

glatard: or just create a `.cache` directory if it doesn't exist? It's also convenient to be able to remove everything under `.cache` without having to worry about loosing configuration.

tools/python/boutiques/searcher.py
glatard: > Aren't tool version and descriptor version the same thing (at least with the way publisher is currently implemented)?

No they are different. The tool is what is described by the descriptor.

glatard: > Also, getting all versions of the tool isn't really possible through the search API. The JSON returned by the search only lists the current version's ID, the number of versions and the original version's ID. Should I include any of those (number of versions or parent ID)?

so is there any way to find the zenodo id of a previous version if we wanted to get back to it? I assume this might require a bit more work, perhaps add an issue on this so that we can get the current implementation, which is already awesome, merged rapidly.

glatard: > The tabular format looks a bit messy on my little VM with all those columns:
> image
> Should I keep it this way or change the format?

Perhaps you could truncate the description and keywords to a max number of chars but I don't find it a very big deal.

erinb90: > > Aren't tool version and descriptor version the same thing (at least with the way publisher is currently implemented)?
> 
> No they are different. The tool is what is described by the descriptor.

That makes sense but currently the publisher sets the metadata as `'version': self.descriptor['tool-version']`. Should this be changed?

tools/python/boutiques/bosh.py
erinb90: Ok, but setting `hide=True` means it won't print `OK`. I guess that's not a big deal.

tools/python/boutiques/searcher.py
glatard: The most important version is the tool version indeed, so that should be the version in Zenodo. Now, what happens if one publishes multiple descriptors with the same tool version and name? How do they differ in Zenodo?

tools/python/boutiques/bosh.py
glatard: It sounds good indeed.

tools/python/boutiques/searcher.py
erinb90: Only the ID will differ. Zenodo allows for multiple records with the same name and version. An example from Sandbox:
image

glatard:
Perhaps add a test for verbose search and we should be good to go!

tools/python/boutiques/searcher.py
erinb90: The search API only returns the latest version of each record, though.

glatard: ok then so be it!

PR #330, opened 2018-11-07
Space fix

glatard:
As reported by @camarasu earlier today. Code was adding the flag separator (default: ' ') even when the input had no flag.

Additional changes:
* Fixed some variable names to snake_case in localExec.
* ExecutorOutput object returned by simulate now contains the generated command lines instead of the empty string. This can be useful in tests.

### DIFF
--- a/tools/python/boutiques/localExec.py
+++ b/tools/python/boutiques/localExec.py
@@ -212,7 +212,7 @@ class LocalExecutor(object):
         of local execution.
         After execution, it checks for output file existence.
         '''
-        command, exit_code, con = self.cmdLine[0], None, self.con or {}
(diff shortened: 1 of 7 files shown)


tools/python/boutiques/localExec.py
```diff
@@ -808,24 +811,24 @@ def escape_string(s):
                         if escape:
                             val = escape_string(val)
                     # Add flags and separator if necessary
-                    if useFlags:
-                        flag = self.safeGet(paramId, 'command-line-flag') or ''
-                        sep = self.safeGet(paramId,
+                    flag = self.safeGet(param_id, 'command-line-flag')
+                    if (use_flags and flag is not None):
```
glatard: this line is the fix (added `flag is not None`)

PR #332, opened 2018-11-08
Minor bug fixes

gkiar:
- fixed validator bug that failed to pass default values of lists even in list type inputs
- made bids apps consistent with creator apps (using square brackets to prevent parameter nesting issues)
- fixed creator for another type of list arguments


glatard:
Looks great!

PR #333, opened 2018-11-10
Inputs can belong to multiple groups

erinb90:
Issue #304 

Removed the restriction that inputs cannot belong to multiple groups. 

Added the following checks to the validator:
* Pairs of mutually exclusive inputs cannot both be part of an all-or-none group and vice-versa, for example:
```
"groups": [
  {
    "id": "group1",
    "name": "group 1 name",
    "members": ["id2", "id3"],
    "mutually-exclusive": true
  },
  {
    "id": "group2",
    "name": "group 2 name",
    "members": ["id1", "id2", "id3"],
    "all-or-none": true
  }
]
```
The above is invalid because `group2` is saying that "id2" and "id3" can be toggled together, while `group1` states that they are mutually exclusive.

* The members of a one-is-required group cannot be a subset of an all-or-none group, for example:
```
"groups": [
  {
    "id": "group1",
    "name": "group 1 name",
    "members": ["id1", "id2", "id3"],
    "all-or-none": true
  },
  {
    "id": "group2",
    "name": "group 2 name",
    "members": ["id1", "id2"],
    "one-is-required": true
  }
]
```
The above is invalid because `group2` states that at least one of "id1" and "id2" must be used, while `group1` states that "id1" and "id2" can both be unused. If `group2` had other members that were not part of `group1` (e.g. "id4") this would be valid because you could use "id4" as the one required and then have all the inputs of `group1` toggled off (someone please confirm if this makes sense).


glatard:
@gkiar, could you have a quick look at this, since you were involved in #304?

PR #336, opened 2018-11-12
Create invocations from argparse parsers

gkiar:
Added a simple method to the creator that creates invocations from parser args, and associated docs and test


gkiar:
@glatard Fixed it! We have to set a `dest` attribute in subparsers in order for the variable to be stored by the parser. This only affects the invocation creator, as valid descriptors are written in either case. I've added a warning accordingly.

PR #346, opened 2018-11-16
Fixes in BIDS importer

glatard:
* BIDS app import now doesn't crash if BIDS app has no 'version' file
* Removed tabs from template


gkiar:
Cool - the first test fails due to indentation and the second due to what appears to be an out of date Zenodo ID for the test. Once those are fixed I'm happy to merge :)

erinb90:
> Cool - the first test fails due to indentation and the second due to what appears to be an out of date Zenodo ID for the test. Once those are fixed I'm happy to merge :)

@gkiar The Zenodo publication test fails sporadically because of their API being unstable. #289 intends to fix this.

gkiar:
@erinb90 yea, I've noticed that, but here's it was an `AssertionError` that two IDs didn't match - is this also caused by the same thing?

glatard:
@gkiar are you ok to merge this now?

gkiar:
Yep! Sorry about the delay.

PR #348, opened 2018-11-16
Add support for JSON objects/Zenodo IDs passed on the CLI in bosh

erinb90:
Issue #277 

Modified the following bosh functions to accept JSON objects and Zenodo IDs for descriptors and/or invocations:
* Test
* Export
* Import

Wrote test cases for accepting JSON objects/Zenodo IDs for all bosh functions. Some test cases were not written:
* Import: no Zenodo test because there are no published descriptors that use schema version 04 to test the `upgrade_04` function
* Export: no Zenodo test because the published "Example Boutiques Tool" has version 0.0.3 and the one used for the export tests has version 0.0.1. I could change it to 0.0.3 everywhere but I figured this would be tedious as we'd just need to change it again every time we publish a new version. 
* Publish: no test with JSON obj because published items must be files; no test with Zenodo since it doesn't make sense to publish an already published descriptor. 
* Invocation: no Zenodo test because the test uses `good.json` which isn't published

### DIFF
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -335,9 +335,7 @@ def test(*params):
     invocation(result.descriptor)
 
     # Extraction of all the invocations defined for the test-cases.
-    with open(result.descriptor) as fhandle:
-        # might just need to be `fhandle` in this context, but not sure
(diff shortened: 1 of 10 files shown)


glatard:
Great work!

PR #350, opened 2018-11-21
Minor changes to bosh search and pull

erinb90:
Fixed a bug in `bosh pull` where if the descriptor isn't in the first 10 search results, it won't be found and the pull will fail. Instead of searching for all boutiques descriptors, it searches with the ZID as the query, which seems to always return the desired descriptor as the first and only result. 

Also added some info messages which I think are necessary even in non-verbose mode:
* `bosh search` prints the number of results displayed out of the total (so the user will know if there are additional results that were not displayed).
* `bosh pull` prints the name of the download folder (the cache directory) so the user will know where to find the pulled file.

### DIFF
--- a/tools/python/boutiques/puller.py
+++ b/tools/python/boutiques/puller.py
@@ -30,7 +30,7 @@ class Puller():
 
     def pull(self):
         from boutiques.searcher import Searcher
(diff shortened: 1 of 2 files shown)


glatard:
It's great that you keep improving this, nice work!

Issue #351, opened 2018-11-21
Cannot run multiple instances of bosh in parallel with a singularity image


prioux:
If invoking bosh multiple time on the same descriptor in the same directory, and the descriptor defines an execution environment in a singularity image, each bosh will attempt to pull and install the singularity image on top of each other. E.g. assuming I want to process some dataset with three different set of parameters:

```shell
bosh exec launch descriptor.json invoke_params1.json &
bosh exec launch descriptor.json invoke_params2.json &
bosh exec launch descriptor.json invoke_params3.json &
```

and assuming descriptor.json has a definition like

```json
    "container-image": {
        "image": "bids/example",
        "index": "docker://",
        "type": "singularity"
    },
```

then all three instances will try to write simultaneously to the file `bids-example.simg` and of course everything will crash and burn and explode and wail in pain.

Two possible solutions:

1) build into bosh the ability to recognize that another bosh is pulling the image and wait for it:

The pull would be performed on a temporary but deterministic name, e.g. `bids-example.simg-downloading` and then renamed at the end, so that other boshes can wait for it.

2) provide a bosh command to separately pull the image and return its path:

```shell
bosh exec prepare descriptor.json
image=bids-example.simg
```

so that we can parse the output of `prepare` and pass the image with `bosh blah blah --simg image`

glatard:
Thanks for reporting this, I think solution 2 is better as it's simpler and the command could be used in other contexts too.

prioux:
It would be even nicer if we could control the location of the image downloaded:

```shell
bosh exec prepare descriptor.json invoke.json --imagepath=/my/cache/location/hello.simg
```

glatard:
Additional notes:

If the image is a Docker image `prepare` should just pull it, so that multiple instances don't step on each other.

`prepare` should be idempotent, i.e., we should be able to run it sequentially as many times as we want without side effects. That way, `bosh exec launch` could invoke `prepare` in place of the current pulling mechanism without worrying about triggering additional transfers.

It would also be nice if `prepare` could be thread-safe to simplify its use: a simple lockfile should do. That would also enable both @prioux's solutions 1. and 2.

prioux:
Yep, all good.

glatard:
Implemented by @erinb90 in just-released 0.5.16 version.

PR #354, opened 2018-11-27
Bosh exec prepare

erinb90:
Issue #351 

Added `bosh exec prepare` which pulls the Docker or Singularity container image for a descriptor and prints the container location. In the case of a Singularity image, this method creates a lockfile for the image file so that multiple threads trying to pull the same image at the same time won't conflict with each other. Also added the option to specify a custom path for the Singularity image.

### DIFF
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -66,8 +66,10 @@ def execute(*params):
                         "set of inputs compliant with invocation schema "
                         "and launches the tool. Simulate: shows sample "
                         "command-lines based on the provided descriptor"
-                        " based on provided or randomly generated "
-                        "inputs.", choices=["launch", "simulate"])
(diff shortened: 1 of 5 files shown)


glatard:
wow, this looks great! Am I right that Docker doesn't need a lock file because it handles concurrent pulls already?

glatard:
I tested it and it seems to be working. I'm happy to merge this as soon as the tests pass. Nice work!

glatard:
fyi, @prioux, we'll release as soon as this is merged.

erinb90:
> wow, this looks great! Am I right that Docker doesn't need a lock file because it handles concurrent pulls already?

@glatard correct, Docker doesn't seem to need a lockfile. I tried pulling the same Docker container simultaneously and it was fine.

I think the Zenodo API was having problems at the time I made this PR because there were a bunch of 500 errors. Can you rerun the Travis build?

Issue #76, opened 2016-10-25
List separator


glatard:
Currently, list elements are separated by spaces on the command line. A custom list separator could be defined.

PR #358, opened 2018-11-29
Added option to specify custom list separator

erinb90:
Issue #76


tools/python/boutiques/localExec.py
```diff
@@ -833,14 +833,16 @@ def escape_string(s):
                     val = in_out_dict[param_id]
                     if type(val) is list:
                         s_val = ""
+                        list_sep = (self.safeGet(param_id, 'list-separator')
+                                    or " ")
```
glatard: I'm not sure I like the `or` here, as the empty string evaluates to `False`. I reckon having fixed a few bugs that were coming from that. I'd say `if list_sep == None: list_sep = " "` is safer.

Issue #327, opened 2018-11-01
Create a common logger


erinb90:
We should have a common logger throughout the project, e.g. for printing Zenodo info and error messages, and possibly other kinds of messages.

PR #364, opened 2018-12-06
Common logger

erinb90:
Issue #327 

Created a common logger to print info, error and warning messages throughout the code. So now, nearly all printed messages will look like this:
```
[ ERROR ] Some error message
[ INFO ] Some info message
[ WARNING ] Some warning message
```


tools/python/boutiques/puller.py
```diff
@@ -11,19 +13,15 @@
     from urllib import urlretrieve
 
 
-class ZenodoError(Exception):
```
glatard: what's the magic here, how can you raise a ZenodoError if the class was never defined?

erinb90: I noticed that `ZenodoError` was defined in 3 places so I removed it from `puller` and `searcher` and imported the one from `publisher`. So no magic there :)

PR #371, opened 2018-12-19
Mock zenodo server

erinb90:
Issue #289

Created mocks for all the Zenodo server calls so that the tests don't randomly fail due to Zenodo's server being weird.

Note: #368 should be merged first so that I can use the mocks for the new tests.

### DIFF
--- a/tools/python/boutiques/tests/test_evaluate.py
+++ b/tools/python/boutiques/tests/test_evaluate.py
@@ -4,6 +4,16 @@ import os
 from unittest import TestCase
 from boutiques import __file__ as bfile
 import boutiques as bosh
+import mock
+from boutiques_mocks import mock_zenodo_search, MockZenodoRecord
+
+
+def mock_get():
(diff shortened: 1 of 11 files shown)


tools/python/setup.py
```diff
@@ -10,7 +10,8 @@
          "termcolor",
          "pyyaml",
          "jsonschema",
-         "tabulate"
+         "tabulate",
+         "mock"
```
glatard: mock seems to be part of the Python library from 3.3. Not sure if it changes anything here.

tools/python/boutiques/tests/test_evaluate.py
```diff
@@ -4,6 +4,16 @@
 from unittest import TestCase
 from boutiques import __file__ as bfile
 import boutiques as bosh
+import mock
```
glatard: +1 for using what seems to be the common way to do it in Python.

tools/python/setup.py
erinb90: It is, but it seems that everything still works fine when it's manually installed.

Issue #362, opened 2018-12-04
Validation error when no value is specified for mandatory inputs that have a default value


erinb90:
The default values should be used, however the validator complains that required inputs are missing. The logic should be rearranged so that the default values are properly substituted in.

Issue #338, opened 2018-11-13
bosh exec launch produces a command line where no optionals have their default


shots47s:
Using the boutiques descriptor at doi 1484547, if an input element is optional and there is a default-value given, the item does not appear on the command line in the temp work script.

So using this descriptor, and this input json:

{
"bids_dir":"d005",
"analysis_level":"participant",
"participant_label":["02"],
"license_file": "/home/users/sbrown/freesurfer/license.txt",
"output_dir":"b2-out"
}

I get the following command line in temp file:
`mkdir -p b2-out ; /run.py d005 b2-out participant --participant_label 01       --license_file /home/users/sbrown/freesurfer/license.txt`

This would imply that optional variables will never get set to defaults and seems like a weird syntax.

glatard:
Currently the default value is only added to mandatory parameters when they have no value. It's easy to change but I'm not sure of the implications of that yet.

shots47s:
Well, in the case of the bids apps, the run.py has the default values, so it is moot.  But I would think it is a matter of the boutique descriptor being a standalone representation of the full command line.  So technically, I guess the value is not "optional" if it has a default, because by definition you are saying the variable has to have a value, and if you don't specify, then this is the value that it needs to be.  So I think that I am wrong in my question and your current default is correct.  Had to have you ask the question to make me realize the answer.  

Thanks for indulging.

Would suggest no change, clear documentation on this matter (as it is not clear from the schema or current docs that this is the case) and closing this issue.

glatard:
Then to make it clear, shall we prevent optional inputs to have a default value?

glatard:
Add a property `internal-default-value` to specify the value that will be used internally for optional inputs when value isn't specified.

erinb90:
@glatard On this topic, I'm wondering if there is a bug in the normal behavior or if I'm just doing something wrong? If I try removing `str_input` from `invocation.json`, it should use the default value which is "aStringValue". But instead I get the following error on `bosh exec simulate`:

```
u'str_input' is a required property

Failed validating 'required' in schema:
    {'$schema': 'http://json-schema.org/draft-04/schema#',
     'additionalProperties': False,
     'allOf': [{'anyOf': [{'required': [u'num_input']},
                          {'required': [u'enum_input']}]}],
     'dependencies': {u'enum_input': {'properties': {u'num_input': {'not': {}}}},
                      u'flag_input': {'anyOf': [{'properties': {u'num_input': {'not': {}}},
[...]

PR #376, opened 2019-01-10
Default values are always used when input values are not provided, fixed validation error

erinb90:
* (#362) Fixed bug where not providing a value for a mandatory input with a default value threw a schema validation error. The default values are temporarily added to the input dictionary right before the invocation is validated to prevent the error. 
* (#338) Default value is always used when a value is not provided, even when the input is optional.
* Minor fixes, including fixing indentation in `invocationSchemaHandler` and fixing the `example1` descriptors to have the proper default value for the `str_input_list` input.

### DIFF
--- a/tools/python/boutiques/localExec.py
+++ b/tools/python/boutiques/localExec.py
@@ -807,16 +807,7 @@ class LocalExecutor(object):
             del self.in_dict[r]
         # Add default values for required parameters,
         # if no value has been given
-        for input in [s for s in self.inputs
-                      if s.get("default-value") is not None
-                      and not s.get("optional")]:
-            if self.in_dict.get(input['id']) is None:
-                df = input.get("default-value")
-                if not input['type'] == 'Flag':
-                    self.in_dict[input['id']] = df
-                else:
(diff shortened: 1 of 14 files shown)


tools/python/boutiques/localExec.py
```diff
@@ -1177,3 +1169,18 @@ def loadJson(userInput, verbose=False):
         return json.loads(userInput)
     except ValueError:
         raise_error(ExecutorError, e)
+
+
+# Adds default values to input dictionary
+# for parameters whose values were not given
+def addDefaultValues(desc_dict, in_dict):
+    inputs = desc_dict['inputs']
+    for input in [s for s in inputs
```
glatard: isn't `input` a reserved keyword?

erinb90: I copied this code over from another place so I didn't think about it, but I looked it up and apparently it's *not* a keyword, but will overwrite the built-in `input` function if used. We don't use `input` anywhere except the publisher, so it doesn't cause a problem, but it's bad practice. I'll change the name.

Issue #377, opened 2019-01-10
bosh is incapable of providing its own version


prioux:
I tried:

```shell
bosh -v
bosh -V
bosh --version
bosh --v
bosh version
```

and all of them just provided the same usage statement. For provenance recording and automatic pipelines, it would be good if at least one of them would print out, on a single line, something like

```
0.5.18
```

so that the invoking context could capture and record that.

glatard:
Thanks for mentioning this @prioux 

Note to whoever will do this: 
* The flag should be `--version` as `-v` is already used for verbose mode.
* The version string should be consistent with the one in `setup.py`, i.e., it should be written only once.

PR #378, opened 2019-01-10
added version to package

gkiar:
I needed something to keep my fingers busy while waiting for a build so did #377 for @prioux.

Usage:

```
$ bosh version
0.5.18
```

In new releases moving forward, we now change the version in the `boutiques/__version__.py` file.

Shout out to the @arokem and the Shablona folks for reminding me the easiest way to do this, [here](https://github.com/uwescience/shablona/blob/master/setup.py#L5-L8)!


prioux:
Er, cool, but given that setup.py already had a VERSION constant defined, wasn't it simpler in bosh.py to just print that out? And if VERSIOn had to be moved to a new file, then couldn't setup.py import VERSION the same way you do in bosh.py, instead of opening the file, and executing() the content?

gkiar:
Explained offline to @prioux - we don't include `VERSION` in `setup.py` because it isn't a part of the package, and we can't import it in `setup.py` like we do in the `bosh.py` script because importing it that way requires the package to be installed (which we can't assume when trying to install the package).

PR #379, opened 2019-01-10
Improved top level help text.

prioux:
This is one example of making the usage statements much better looking. It only modifies the top level usage statement; similar modifications could be made for the usage statements of each subcommand.

I changed the very first line of the usage statement from the nonsensical:

```
The tool within boutiques/bosh you wish to run.
```

to

```
Provide one of these subcommands:
```

Speaking of the subcommands, they also require some rewriting of their helps, because when we type (e.g.) `bosh create -h` we do not get back a usage statement that says `bosh create` in it, instead we get `Boutiques descriptor creator` in its place:

```
usage: Boutiques descriptor creator [-h] [--docker-image DOCKER_IMAGE]
                                    [--use-singularity]
                                    descriptor

positional arguments:
  descriptor            Output file to store descriptor in.
(...)
```

### DIFF
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -455,26 +455,37 @@ def pull(*params):
 
 def bosh(args=None):
     parser = ArgumentParser(description="Driver for Bosh functions",
-                            add_help=False)
+                            add_help=False, formatter_class=RawTextHelpFormatter)
     parser.add_argument("function", action="store", nargs="?",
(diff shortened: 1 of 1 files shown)


prioux:
I'll wait one day and then close the PR.

prioux:
Huh, I forgot to return to this. Will clean this week (either by closing the PR or by reformating the text)

prioux:
I don't have time to rebase and fix the layout, so I'll close this.

PR #381, opened 2019-01-11
Removed some redundant/ambigious code. Function is the same, but there is less code and intent is clearer

ramou:
* I've localised the loadJson call on result.invocation in invocation() to simplify the code.
 * I've in-lined the json String in test's call of invocation instead of exporting a json file and passing that.


glatard:
Closing this as it was open against `master` instead of `develop`.

Issue #384, opened 2019-01-14
Create a README for setting up local dev environment


erinb90:
We should have a document for new Boutiques developers on how to set up their local environment, showing which commands to run, what needs to be installed, etc.

PR #389, opened 2019-01-21
Readme for setting up local dev environment

erinb90:
#384

This is just based off my memory of when I first set up my environment so it's possible I am missing things. Different systems might have different setup requirements, but I tried to make this guide as general as possible.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -3,46 +3,51 @@
 This guide will help you set up your local environment for coding and testing Boutiques.
 
 ## 0. Things you'll need
-- **A Linux system**. It _may_ be possible to develop Boutiques on Windows, but will likely cause many headaches.
-- **Root (sudo) access**. Unless your system already has everything you need installed, you'll need root permission to install new programs.
+- **A Linux shell**. It _may_ be possible to develop Boutiques on Windows, but will likely cause many headaches (if you successfully develop for Boutiques on Windows, please update this guide with details!).
+- **Git**. If you don't have Git, instructions to install it are [here](https://git-scm.com/download/linux).
 - **Python**. Boutiques supports any version >= 2.7. It's a good idea to have multiple versions of Python installed (e.g. 2.7 and 3.7) so you can test your code for version compatibility.
 - **pip**. Most versions of Python will come with it, but in case you don't have it, instructions to install it are [here](https://pip.pypa.io/en/stable/installing/).
-- **Docker**. If you don't already have Docker, follow the [instructions](https://docs.docker.com/install/overview/) to install Docker CE.
-- **Singularity**. Follow the instructions [here](https://singularity.lbl.gov/install-linux) to download the latest stable release of Singularity.
-- **Git**. If you don't have Git, instructions to install it are [here](https://git-scm.com/download/linux).
+- **Docker or Singularity**. Depending on the tools you plan to work with and where they are installed, you'll want to have either Docker or Singularity installed (ideally both).
+  - To install Docker, follow the  [instructions](https://docs.docker.com/install/overview/) to install Docker CE.
+  - To install Singularity, follow the [instructions](https://singularity.lbl.gov/install-linux) to download the latest stable release of Singularity.
+- **Root (sudo) access**. If your system doesn't already have all of the above installed, you'll need root permission to install new programs.
 
 ## 1. Clone the boutiques repository
 - Fork the [Boutiques repository](https://github.com/boutiques/boutiques).
 - Clone your fork locally: 
-    `git clone https://github.com/<YOUR-GITHUB-USERNAME>/boutiques.git`
+    - `git clone https://github.com/<YOUR-GITHUB-USERNAME>/boutiques.git`
+    - Or, if you installed your SSH keys on GitHub:  `git@github.com:boutiques/boutiques.git`
 
 ## 2. Create a virtual environment
 - Install virtualenv:
-  `sudo pip install virtualenv`
+  - `pip install --user virtualenv`
 - Create a virtual envrionment in the Python tools root directory:
-  `cd boutiques/tools/python`
-  `virtualenv env`
+  - `cd boutiques/tools/python`
+  - `mkdir env`
+  - `virtualenv env/boutiques -p python3.6`  (replace python3.6 with the executable for the version of Python you wish to use)
 - Use the environment whenever you work on Boutiques by running:
-  `source env/bin/activate`
+  - `source env/boutiques/bin/activate`
 - Note: if you wish to leave the virtual environment, simply run `deactivate`.
 
 ## 3. Install Boutiques, test framework and validation tools
 From within your virtual environment in the `boutiques/tools/python` directory:
 - Upgrade setuptools: `pip install --upgrade setuptools`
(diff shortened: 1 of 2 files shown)


GETTING_STARTED.md
```diff
@@ -0,0 +1,63 @@
+# Getting Started as a Boutiques Developer
+
+This guide will help you set up your local environment for coding and testing Boutiques.
+
+## 0. Things you'll need
+- **A Linux system**. It _may_ be possible to develop Boutiques on Windows, but will likely cause many headaches.
```
gkiar: Maybe update this to say a **Linux shell**, since Mac works. Also potentially add a parenthetical at the end of the sentence that is "(if you successfully develop for Boutiques on Windows, please update this guide with details!)"

GETTING_STARTED.md
```diff
@@ -0,0 +1,63 @@
+# Getting Started as a Boutiques Developer
+
+This guide will help you set up your local environment for coding and testing Boutiques.
+
+## 0. Things you'll need
+- **A Linux system**. It _may_ be possible to develop Boutiques on Windows, but will likely cause many headaches.
+- **Root (sudo) access**. Unless your system already has everything you need installed, you'll need root permission to install new programs.
```
gkiar: I'd put this at the bottom of the list, since it only is required if other things are not installed. I develop on Compute Canada for Boutiques, and I certainly don't have root there :)

GETTING_STARTED.md
```diff
@@ -0,0 +1,63 @@
...
+- **Git**. If you don't have Git, instructions to install it are [here](https://git-scm.com/download/linux).
+
+## 1. Clone the boutiques repository
+- Fork the [Boutiques repository](https://github.com/boutiques/boutiques).
+- Clone your fork locally: 
+    `git clone https://github.com/<YOUR-GITHUB-USERNAME>/boutiques.git`
+
+## 2. Create a virtual environment
+- Install virtualenv:
+  `sudo pip install virtualenv`
```
gkiar: I'd probably recommend `pip install --user virtualenv` rather than suggest they use `sudo`, here.

GETTING_STARTED.md
```diff
@@ -0,0 +1,63 @@
...
+- Create a virtual envrionment in the Python tools root directory:
+  `cd boutiques/tools/python`
+  `virtualenv env`
```
gkiar: Personally, I like having clearly-named environments, so would recommend the following two lines in replacement of the one I'm commenting on:
```
mkdir env
virtualenv env/boutiques -p python3.6  # replace python3.6 with the executable for the version of Python you wish to use
```

GETTING_STARTED.md
```diff
@@ -0,0 +1,63 @@
...
+- Use the environment whenever you work on Boutiques by running:
+  `source env/bin/activate`
```
gkiar: Given the above comment, this line would switch to `source env/boutiques/bin/activate`.

GETTING_STARTED.md
```diff
@@ -0,0 +1,63 @@
...
+- Note: if you wish to leave the virtual environment, simply run `deactivate`.
+
+## 3. Install Boutiques, test framework and validation tools
+From within your virtual environment in the `boutiques/tools/python` directory:
+- Upgrade setuptools: `pip install --upgrade setuptools`
+- Install pytest and pycodestyle: `pip install pytest pycodestyle`
```
gkiar: They'll probably also need `pytest-runner`, no? Maybe that's only used by `coverage`...

GETTING_STARTED.md
```diff
@@ -0,0 +1,63 @@
...
+  - pytest is a unit testing framework for Python.
+  - pycodestyle checks if your code conforms to Python style conventions.
+- Install jsonlint: `npm install jsonlint -g` 
+  - In case you don't have Node installed, instructions to install it are [here](https://www.npmjs.com/get-npm).
+  - jsonlint is a tool to validate JSON files. 
+- Install Boutiques: `pip install boutiques` or just `pip install .`
```
gkiar: I'd suggest we recommend `pip install -e .` for installing in a dev environment. Advantage over `pip install boutiques` is obvious (you get the dev version), and over `.` without the `-e` is that the `-e` switch enables "editable mode", which automatically updates the library whenever you make changes to files.

GETTING_STARTED.md
```diff
@@ -0,0 +1,63 @@
...
+## 5. Try it out!
+- Run `pytest` in the `boutiques/tools/python` directory to run all tests.
+- Run an individual test file with `pytest boutiques/tests/test_<something>.py`
+- Validate your code style by running:
+  `pycodestyle --max-line-length=80 boutiques/*.py boutiques/tests/*.py`
+- Validate changes to the Boutiques JSON schema by running:
+  `jsonlint -q boutiques/schema/descriptor.schema.json`
+
+## Things to Note
+- Whenever you make changes to the code and wish to test the `bosh` command-line interface, you must reinstall Boutiques (`pip install boutiques` or `pip install .`). Otherwise, you will be running Boutiques with the old code.
```
gkiar: This note is fixed by installing boutiques with `-e .`.

glatard:
That's super useful @erinb90! See my comments.

GETTING_STARTED.md
```diff
@@ -0,0 +1,63 @@
...
+
+This guide will help you set up your local environment for coding and testing Boutiques.
+
+## 0. Things you'll need
+- **A Linux system**. It _may_ be possible to develop Boutiques on Windows, but will likely cause many headaches.
+- **Root (sudo) access**. Unless your system already has everything you need installed, you'll need root permission to install new programs.
+- **Python**. Boutiques supports any version >= 2.7. It's a good idea to have multiple versions of Python installed (e.g. 2.7 and 3.7) so you can test your code for version compatibility.
+- **pip**. Most versions of Python will come with it, but in case you don't have it, instructions to install it are [here](https://pip.pypa.io/en/stable/installing/).
+- **Docker**. If you don't already have Docker, follow the [instructions](https://docs.docker.com/install/overview/) to install Docker CE.
+- **Singularity**. Follow the instructions [here](https://singularity.lbl.gov/install-linux) to download the latest stable release of Singularity.
```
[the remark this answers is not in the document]
gkiar: Yeah, maybe say "Docker or Singularity" here, because it depends greatly on the system you're using and _generally_ if it works for one it'll work for both, unless you're editing that specific section of code.

GETTING_STARTED.md
erinb90: I totally didn't realize that `pip install boutiques` will just install the latest release, oops

ramou: gold

GETTING_STARTED.md
gkiar: I think `coverage` uses it when it's calling `pytest` as we do in the `.travis.yml` (which is how I run tests locally). Could you try running the `.travis.yml` test command on your environment, and see if you need `pytest-runner` for that?

The command is, I believe `coverage run --source boutiques setup.py pytest`, run from the `tools/python` directory.

erinb90:
This PR should be ready unless anyone has other comments. Not sure why that build error happened, might need to run a new one.

gkiar:
Hey @erinb90 - yep, looks great, thanks so much for making this!

PR #403, opened 2019-02-09
Fixed codestyle

erinb90:
Fixed the overindentation that the new version of pycodestyle was complaining about.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ python:
 - 3.6-dev
 - 3.7-dev
 - nightly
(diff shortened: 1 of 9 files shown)


glatard:
Hi @erinb90, shall we just run our tests with PyPy >= 5.4 and clear this issue?

erinb90:
@glatard Yes I think we should.

glatard:
perfect, would you like to do it as part of this PR?

erinb90:
Can we merge this?

PR #406, opened 2019-02-11
Utf8fix

ramou:
- I have xmlcharrefreplaced strings coming in from Zenodo as some of them could contain unicode data, as we saw with the author name in the BEst tool. 
   - Minimal code to see the error: `print(u'\xef')`
   - Terminal support for unicode is not reliable. On mac, one likely has UTF-8 support, which one can check via `import sys` and `print(sys.stdout.encoding)` but one can't be reliably sure elsewhere.
   - To solve, we encode the unicode as an ascii string and unicode is replaced with a text notation: `&#239`; replaces `\xef` which is the encoding for &#239;.
   - Without this utf-8 support, bosh will give error messages in non-utf-8 terminals and tests will not pass.
   - This solution was messy as support to check for unicode seems to vary in a non-backwards-compatible way across different python versions. To solve this, in versions prior to 3 we can test if the string is of type unicode. After version 3, the unicode type ceases to exist and all strings are treated as type string, so we just need to check for that. However, in some version of python 3, this encoding is to bytes, and is incompatible with later string interaction, thus we further decode our encoded python3 bytes so it behaves uniformly for all of python3 (converts things back to a plain ascii string). This change was placed after the OrderedDict was built in searcher.py, but before truncation was performed as our encoding lengthens strings and we want things to still pretty print nicely.
   - In theory we could wrap this in a check for UTF-8 terminal support and only apply this ugly solution when it isn't found
   - This fixes https://github.com/boutiques/boutiques/issues/386
 - I have also considered the idea of checking for utf-8 support in the console, I'm just not sure how reliable that would be. However, we could at least make it look as it did on previous known utf-8 supporting consoles (mac) and still have the xmlref fix for everyone else so it wouldn't crash.


ramou:
Needs to get updated with the current state of devel before closer consideration.

glatard:
Discussion from the meeting: will check if console supports UTF8. If yes, keep the current behaviour, otherwise, replace UTF8 chars with XML escapes.

ramou:
I can force a terminal to be all utf-8-y
```
locale-gen en_US.UTF-8
export LANG=en_US.UTF-8 LANGUAGE=en_US.en LC_ALL=en_US.UTF-8
```

Issue #405, opened 2019-02-10
The semantics of bosh --imagepath is wrong


prioux:
The bosh subcommands `prepare` and `launch` both support the option `--imagepath`.

It is terrible to use.

First, counter intuitively, the option is not the path to an image file, it is that path to the directory where that image is stored. It should be called 'imagedir'!

Second, when an external system wants to prepare an execution, it makes it impossible to find the name of the image file without guessing. One has to first, make up a list of the files in the target directory, then invoke the prepare command, then read again the target directory to find out the name of the image file. And that is, assuming that in the meantime only one new file will have been created and it happens to be the image. Of course the invoker could parse the message that bosh prints out when it's done, but that message is a) for humans, and likely to change 2) is not the same with the prepare command has pulled the image vs when the image was already found in the directory.

The proper behavior would have been for the `--imagepath` option to specify the path to the image file itself, chosen by the invoker. That allows the invoker to give it a name of its own choosing, to verify that the image was pulled properly, and even to override the image specified in the descriptor with another image if necessary.

prioux:
Greg tells me that for ‘launch’ it is the path to the image, that’s good. Only  ‘prepare’ needs to be fixed.

erinb90:
I made the option be the directory name because I didn't realize you could specify a different image name when pulling an image, but now I see there is a `--name` option for `singularity pull` that lets you specify the image name. So I think we can use that to pull the image with any name we want.

erinb90:
Also, to set the singularity pull folder you need to supply the directory name, so the simplest thing was just to have the user supply that. But it should be easy enough to have the user supply the full path and get the directory name from there.

shots47s:
So.. is the solution to make --imagepath specify a specific image or to make --imagedir and a separate flag that will provide the name of the singularity image? It is advantageous for us to have the prior as we would like to pull and cache the singularity image so that we do not need to pull it every time we run something and do not need to do extra parsing to find the name of the image.

glatard:
So let's fix prepare so that `--imagepath` takes the path of the image instead of the directory.

prioux:
There is still abug when the path is a plain basename (discussed in https://github.com/boutiques/boutiques/issues/404#issuecomment-464502819 ).Maybe the comments I made over there should have been here.

glatard:
Closing this, leaving #404 open to track the path issues for lockdir creation.

Issue #404, opened 2019-02-10
Running several 'bosh prepare' launched from distinct CWDs crash and leave files behind


prioux:
This bug report applies to the 'develop' branch, currently at 9d4cd352e70aaed8ea2209c268c65ec443380e57 .

Given a Boutiques descriptors that requires a Singularity container, if I try to:

* run several 'bosh prepare' in parallel
* where each bosh prepare is invoked from a distinct current working directory
* where the target directory --imagepath for the image is the same for all of them

then:

* while the `bosh prepare` are running, several image file will be created an pulled in parallel in the target directory (with distinct temp extensions)
* then all `bosh prepare` will crash with a python stack trace
* the target directory for the pull will finally contain a single spurious `.tmp` copy of the image file

Here is a shell session that demonstrate the problem:

```
unix> ls
boutiques_adjusted.json
unix> mkdir -p t/cwd_a t/cwd_b t/pull
unix> cd t/cwd_a
unix> bosh exec prepare --imagepath ../pull ../../boutiques_adjusted.json &
[1] 28466
unix> ls -l
total 12
drwxr-xr-x 2 prioux prioux 4096 Feb 10 13:53 shots47s-bids-fmriprep-1.2.3.simg-lock/
unix> cd ../cwd_b
unix> bosh exec prepare --imagepath ../pull ../../boutiques_adjusted.json &
[2] 28596
unix> ls -l
total 12
drwxr-xr-x 2 prioux prioux 4096 Feb 10 13:54 shots47s-bids-fmriprep-1.2.3.simg-lock/
unix> ls -l ../pull
total 1720332
-rw-r--r-- 1 prioux prioux 1525678080 Feb 10 13:54 shots47s-bids-fmriprep-1.2.3.simg.tmp.XIKbdB
-rw-r--r-- 1 prioux prioux  235929600 Feb 10 13:54 shots47s-bids-fmriprep-1.2.3.simg.tmp.vKYr_D
(time passes...)
unix> Traceback (most recent call last):
  File "/home/users/prioux/.local/bin/bosh", line 9, in <module>
    load_entry_point('boutiques==0.5.18', 'console_scripts', 'bosh')()
  File "/home/users/prioux/src/boutiques/tools/python/boutiques/bosh.py", line 523, in bosh
    out = execute(*params)
  File "/home/users/prioux/src/boutiques/tools/python/boutiques/bosh.py", line 211, in execute
    container_location = executor.prepare()[1]
[...]

PR #409, opened 2019-02-12
Fixed imagePath for bosh exec launch/prepare

erinb90:
Fixes for #404, #405 

* Fixed the `--imagepath` option in `bosh exec launch` and `bosh exec prepare` to be the absolute path of the Singularity image (this can refer to either an existing image file or the path you want the image to have when it's pulled)
* Made it so the lockdir is created in the same location as the imagepath, so that processes can be run from different locations using the same imagepath and it should not cause problems.
* Removed the `finally` clause that deletes the lockdir and pullfolder environment variable and replaced it with a method that does this only once a container image has been found locally or pulled successfully. The `finally` clause was causing a bug where if a process timed out while trying to pull the image, it would still delete the lockdir, causing issues for other processes that are still running. 

@prioux

### DIFF
--- a/tools/python/boutiques/localExec.py
+++ b/tools/python/boutiques/localExec.py
@@ -439,14 +439,18 @@ class LocalExecutor(object):
                 except OSError:
                     time.sleep(5)
                 else:
-                    # Check if container was created while waiting
-                    if self._singConExists(conName, imageDir):
+                    try:
+                        # Check if container was created while waiting
+                        if self._singConExists(conName, imageDir):
+                            conPath = op.abspath(op.join(imageDir, conName))
+                            container_location = "Local ({0})".format(conName)
+                        # Container still does not exist, so pull it
+                        else:
+                            conPath, container_location = self._pullSingImage(
+                                conName, conIndex, conImage, imageDir, lockDir)
+                        return conPath, container_location
+                    finally:
                         self._cleanUpAfterSingPull(lockDir)
-                        conPath = op.abspath(op.join(imageDir, conName))
-                        return conPath, "Local ({0})".format(conName)
-                    # Container still does not exist, so pull it
-                    return self._pullSingImage(conName, conIndex, conImage,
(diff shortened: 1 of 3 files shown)


tools/python/boutiques/localExec.py
```diff
@@ -474,15 +483,23 @@ def _pullSingImage(self, conName, conIndex, conImage):
         sing_command = "singularity pull --name " + pull_loc
         (stdout, stderr), return_code = self._localExecute(
                                                 sing_command)
+        self._cleanUpAfterSingPull(lockDir)
```
glatard: Style nitpick: it would be clearer if all the calls to `cleanUpAfterSingPull` would be in the while loop above.

tools/python/boutiques/localExec.py
```diff
@@ -411,56 +410,66 @@ def prepare(self):
...
+                    # Container still does not exist, so pull it
+                    return self._pullSingImage(conName, conIndex, conImage,
+                                               imageDir, lockDir)
+            # If loop times out, check again for existence, otherwise
+            # raise an error
+            if self._singConExists(conName, imageDir):
+                conPath = op.abspath(op.join(imageDir, conName))
+                return conPath, "Local ({0})".format(conName)
+            raise_error(ExecutorError, "Unable to retrieve Singularity "
+                        "image.")
```
glatard: Could you make sure that all the lines in this loop are covered by tests?

erinb90: Ok, it's a bit tricky since we can't really have 2 processes running in parallel in a test, but I think I can do it with mocks.

erinb90:
What happened on Travis? It says it couldn't download Python.

erinb90:
Noticed there's still a few lines not covered, so I'll add a couple more tests soon.

glatard:
Coveralls is crashing all PRs atm, but you also have a pycodestyle error in there.

Issue #393, opened 2019-02-05
Explicit version consolidation


glatard:
Detecting existing versions based on tool name only isn't very robust, as one may want to update the tool name or correct typos in it. Instead, we should provided a `--replace <zenodo_id>` option in `bosh publish` that would consolidate the new publication with an existing one.

The existing name-based matching can be kept too, as it is optional anyway. The new one would provide a way to make sure that the record will be updated.

erinb90:
So I wrote some code for this that I am pretty certain is good, but every time I try to publish an updated version of something on Sandbox I get a 400 error when trying to update the metadata of the new deposition version. It complains about not being allowed to update the doi, but I am not sending the doi in the data. I tried it without my changes and got the same thing. So I am really confused because it used to work fine. I sent a message to Zenodo support asking if they had any idea. If someone wants to try it on their machine and see if they get the same error please do. Just publish any descriptor to Sandbox, remove its doi and try to publish a new version.

erinb90:
After some investigation I figured out that they recently changed their API such that it now requires you to send the doi when updating a deposition. If you don't send it, it will default to `''`, so it will think you are trying to change the doi to `''`, hence the error message saying I can't change the doi. I commented on their GitHub saying they should make this clear in the API docs.

glatard:
@erinb90, does it mean that @shots47s' current issues with updating descriptors are coming from this API change, i.e., updating descriptors currently doesn't work?

PR #410, opened 2019-02-15
Explicit version consolidation

erinb90:
#393 

Added a `--replace <zenodo-id>` option to `bosh publish` to publish updated versions using the unique Zenodo ID.

### DIFF
--- a/tools/python/boutiques/publisher.py
+++ b/tools/python/boutiques/publisher.py
@@ -147,7 +147,7 @@ class Publisher():
             print_info("Deposition of new version succeeded", r)
         new_url = r.json()['links']['latest_draft']
         new_zid = new_url.split("/")[-1]
-        self.zenodo_update_metadata(new_zid)
+        self.zenodo_update_metadata(new_zid, r.json()['doi'])
         self.zenodo_delete_files(new_zid, r.json()["files"])
         return new_zid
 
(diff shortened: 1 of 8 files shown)


gkiar:
Ouh, I need to do exactly this, this morning! I'll quickly test it out :)

gkiar:
Quick request: can we separate `--replace` into two options, such as `--replace` and `--zenodoID`?

Use case: if the descriptor already has a DOI and I want to update it, I just want to type `--replace` and have it know to a) remove the existing DOI from the descriptor, b) replace that tool, and c) add the new DOI back.

erinb90:
@gkiar Ohhh good point. Yeah I'll do that :)

glatard:
This looks ready-to-merge to me, great work! @gkiar, any last word?

tools/python/boutiques/publisher.py
```diff
@@ -15,19 +15,26 @@ class Publisher():
 
     def __init__(self, descriptor_file_name,
                  verbose, sandbox, no_int,
-                 auth_token):
+                 auth_token, replace, id):
```
gkiar: Should we give `replace` and `id` default values, to make their behaviour more clear? I understand that `replace=False` and `id=None` are the defaults. I realize this is only being called internally, so isn't *strictly* needed, it just increases the readability of code in modular pieces, in my opinion.

gkiar: Perhaps this also means doing the same for other params, `sandbox=False`, `verbose=False`, `no_int=False`, and bumping `auth_token` up in the list since it's required.

erinb90: Good idea. That way if we need to create a Publisher instance anywhere else in the code, we won't need to provide values for every single argument. The same should be done for other classes that have optional values.

gkiar: Exactly - we can create an issue go to back and do this for more, but want to just make that adjustment here for now?

PR #412, opened 2019-02-18
Boolean command-line options fix.

prioux:
When 'optional' was set to true for an input
flag (e.g. "-x") and the input is NOT specified in
the invoke.json, then the command line built got
the flag anyway (e.g "-x" was added).

There is no issue associated with this bug.

### DIFF
--- a/tools/python/boutiques/localExec.py
+++ b/tools/python/boutiques/localExec.py
@@ -1196,7 +1196,8 @@ def loadJson(userInput, verbose=False):
 def addDefaultValues(desc_dict, in_dict):
     inputs = desc_dict['inputs']
     for in_param in [s for s in inputs
-                     if s.get("default-value") is not None]:
+                     if s.get("default-value") is not None
+                     and not s.get("optional")]:
         if in_dict.get(in_param['id']) is None:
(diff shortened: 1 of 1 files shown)


glatard:
Thanks for this @prioux. This error only occurs when the input has a default value, right?

glatard:
Hi @prioux, we had some discussion about this in #338 (and in meetings) and we decided that the desired behaviour was to add default values to optional inputs that don't have a value. Otherwise, default values associated with optional inputs would never be used and that would only be misleading. #338 was fixed in #376, together with another bug. Your changes revert what was done in #376. 

So:
- Does your descriptor have a default value for the problematic optional input?
- If yes, do you agree that this default value is useless and that the right fix is to just remove it?

prioux:
I am creating a bug report about this. Just wait until I provide it and let me know what you think. There is a real bug int here.

prioux:
Here we go, it's at #415, I think you will see there is plenty of evidence that something is wrong.

prioux:
Closing; the main devs will produce a better patch that will not re-introduce behaviors that were meant to be phased out (see #415).

Issue #398, opened 2019-02-07
Make the Python API works on Windows


glatard:
- [ ] hunt for '/' and '\n'
- [ ] setup a testing framework for Windows
@erinb90 and @ramou have tried it. It would be useful in particular for integration in SPM (#391).

ramou:
To clarify, my IDE is in Windows, but I test in a Ubuntu Trusty container sitting on a Debian Stretch server.

PR #413, opened 2019-02-19
Fixed bug where specifying only the basename for imagepath failed

erinb90:
#404 #405 
@prioux

### DIFF
--- a/tools/python/boutiques/localExec.py
+++ b/tools/python/boutiques/localExec.py
@@ -413,7 +413,7 @@ class LocalExecutor(object):
 
             if self.imagePath:
                 conName = op.basename(self.imagePath)
-                imageDir = op.dirname(self.imagePath)
+                imageDir = op.dirname(self.imagePath) or "./"
             else:
                 conName = conImage.replace("/", "-").replace(":", "-") + ".simg"
                 imageDir = "./"
--- a/tools/python/boutiques/localExec.py
+++ b/tools/python/boutiques/localExec.py
@@ -413,10 +413,10 @@ class LocalExecutor(object):
 
             if self.imagePath:
                 conName = op.basename(self.imagePath)
-                imageDir = op.dirname(self.imagePath) or "./"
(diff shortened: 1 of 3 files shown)


tools/python/boutiques/localExec.py
```diff
@@ -413,7 +413,7 @@ def prepare(self):
 
             if self.imagePath:
                 conName = op.basename(self.imagePath)
-                imageDir = op.dirname(self.imagePath)
+                imageDir = op.dirname(self.imagePath) or "./"
```
glatard: Not sure if this is portable: replace with `os.getcwd()`?

prioux: The dirname of a plain basename is `.`, so it's fine.

A comment though. I'd not use the trailing `/`, that's bad semantics. Accessing directories with a trailing slash has different behavior on different unix system. The name of the dir is really `.`.

gkiar: But on Windows this may not work, right @prioux ? We haven't yet adding testing to Windows, but that is the goal, so `os.getcwd()` would have a higher likelihood of being robust to the switch, I think?

prioux: I think windows support "." to mean the current directory... otherwise how do you even specify the current working directory?

prioux: That being said, I'm fine with this PR.

prioux: There's already #398 for windows compatibility anyway. And lots of stuff in here will break way before all the path operations are adjusted (e.g. popen())

prioux: I fond something that will please everyone!

`os.path.normpath("")` returns "." and will probably work OK on windows too (whatever the convention over there is).

So the code above could be written as

```python
imageDir = op.normpath(op.dirname(self.imagePath))
```

prioux: (Tested in interactive python 2.7.5)

Issue #359, opened 2018-12-03
Add a property for online platform(s) having the tool installed


glatard:
Add a property `online-platforms` containing a list of URLs where the descriptor is installed and available. Upon publication, the elements in this list should be added to the 'hasPart' list in Zenodo. That would allow users to be pointed to CBRAIN, VIP and other platforms where the descriptor can be found. As discussed with @shots47s during the CONP meeting today.

glatard:
This is now critical to the integration of Boutiques tools in CONP. We'd like to be able to leverage that for the April 15th CONP prototype.

paiva:
@glatard why is this issue now closed?

PR #416, opened 2019-02-19
Online platforms

glatard:
#359 

Descriptors can now have an array of URLs where the tool is installed. On publication, URLs are published through relation 'hasPart'.

### DIFF
--- a/tools/python/boutiques/schema/descriptor.schema.json
+++ b/tools/python/boutiques/schema/descriptor.schema.json
@@ -516,6 +516,15 @@
                 ]
             }
         },
+        "online-platform-urls": {
+            "id": "http://github.com/boutiques/boutiques-schema/online-platforms",
+            "type": "array",
+            "description": "Online platform URLs from which the tool can be executed.",
+            "items": {
+                "type": "string",
+                "pattern": "^https?://"
+            }
+        },
         "output-files": {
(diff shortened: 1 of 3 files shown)


tools/python/boutiques/schema/descriptor.schema.json
```diff
@@ -516,6 +516,15 @@
                 ]
             }
         },
+        "online-platform-urls": {
+            "id": "http://github.com/boutiques/boutiques-schema/online-platforms",
+            "type": "array",
+            "description": "Online platform URLs from which the tool can be executed.",
+            "items": {
+                "type": "string",
+                "pattern": "^https?://"
```
prioux: Question. Does the schema validator applies the regex using multiline mode? If so, it would allow a descriptor to contain a string such as `"bouahaha\nhttps://example.com/"` and still calls it 'valid', because the caret would match the zero-width character after the newline.

glatard: It looks fine:
`
[ ERROR ] 'bouahaha\nhttps://example.com/' does not match '^https?://'

Failed validating 'pattern' in schema['properties']['online-platform-urls']['items']:
    {'pattern': '^https?://', 'type': 'string'}

On instance['online-platform-urls'][2]:
    'bouahaha\nhttps://example.com/'
`

gkiar:
Looks good to me, @glatard

Issue #415, opened 2019-02-19
Default values for flag cause crash and command line changes


prioux:
This bug report provide evidence for the PR #412.

First, test files. I provide here four files called `b1.json`, `i1.json`, `i2.json`, and `i3.json`:

* b1.json

```json
{
    "name": "BinTrue", 
    "tool-version": "1.0.0.0.0.0",
    "description": "The True Executable", 
    "command-line": "/bin/true [DASH_T] [DASH_F]", 
    "author": "Pierre Rioux",
    "url": "https://github.com/prioux",
    "schema-version": "0.5", 
    "inputs": [
        {
            "command-line-flag": "-t", 
            "description": "A flag with default true",
            "id": "dash_t", 
            "name": "dash_t", 
            "optional": true, 
            "type": "Flag", 
            "value-key": "[DASH_T]",
            "default-value": true
        },
        {
            "command-line-flag": "-f", 
            "description": "A flag with default false",
            "id": "dash_f", 
            "name": "dash_f", 
            "optional": true, 
            "type": "Flag", 
            "value-key": "[DASH_F]",
            "default-value": false
        }
    ], 
    "tags": {
        "application-type": [
            "unix"
        ], 
        "domain": [
            "bosh", 
            "bash"
        ]
    } 
}
```

* i1.json

```json
{
  "dash_t": true,
  "dash_f": true
}
```

* i2.json

```json
{
  "dash_t": false,
  "dash_f": false
}
```

* i3.json

```json
{
}
```

Now for the behaviors. Using the simulate command we get:

* In branch `master`

```
master boutiques > bosh exec simulate b1.json -i i1.json 
Generated Command:
/bin/true -t -f
master boutiques > bosh exec simulate b1.json -i i2.json 
Generated Command:
/bin/true  
master boutiques > bosh exec simulate b1.json -i i3.json 
Generated Command:
/bin/true  
```

* In branch `develop`

```
develop boutiques > bosh exec simulate b1.json -i i1.json 
Generated Command:
/bin/true -t -f
develop boutiques > bosh exec simulate b1.json -i i2.json 
Generated Command:
/bin/true -t -f
develop boutiques > bosh exec simulate b1.json -i i3.json 
[ ERROR ] 'False' is not of type 'boolean'

Failed validating 'type' in schema['properties'][u'dash_f']:
    {'type': 'boolean'}

On instance[u'dash_f']:
    'False'
```

[...]

PR #417, opened 2019-02-20
Fixed handling of inputs of type 'Flag'

glatard:
#415 
#412 

* Default values are now properly handled for inputs of type 'Flag'.
* Added tests based on the examples in #415.
* Internal representation of values of 'Flag' inputs is now boolean instead of string. This was causing inconsistencies and unnecessary casts. A consequence is that passing strings as Flag values in the invocations, e.g., `"myflag": "false"`, won't work any more. Instead, invocations will have to use JSON's boolean type: `"myflag": false`. I don't think we want to support passing strings to activate flags, and I don't think anyone critically relies on that at the moment.

@prioux

@gkiar or @erinb90: I'd like a second pair of eyes on this, if you have time.


prioux:
Seems all reasonable to me. That is, until I find all my descriptors breaking! :-)

(But I'll fix the descriptors if that happens, or communicate with the owners)

prioux:
Just so you know, I have 40 or so descriptors in CBRAIN, and I searched for the strings `"true"` and `"false"` in them. I found only just a few, and none of them were used for `Flag` inputs. So all is good.

PR #441, opened 2019-03-10
Travis CI is ignoring the jsonschema requirement, this forces it

### LINKED ISSUE #439: Travis CI is no longer working
A of 0.5.19 Travis CI integration appears not to pass tests for python 2.7 and 3.6. I've checked a lot of stuff and I'm almost certain the problem is with Travis, or something that is getting installed during the builds. Nonetheless, tests aren't passing and that's a big nuisance.

@gkiar has mentioned an interest in Circle CI (it has some good reviews) given the fear that Travis might disappear given news about layoffs and them being bought. I can't tell if jumping ship is the way to go, or if there's a simpler solution to resolve this, but it'd be nice if it were working again.

The only thing I note is that it's using `subprocess` when it's getting errors, but stuff like `bosh launch` must be getting used elsewhere without tests failing, so that's pretty tenuous.

ramou:
This closes #439

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ before_install:
 
 install:
 - pip install --upgrade setuptools
+- pip install "jsonschema>=2.6.0,<3.0.0"
 - pip install pytest pytest-runner pycodestyle coveralls
 - pip install .
 - pip freeze


glatard:
Thanks for sorting this out @ramou! Is there a good reason for us to not use the latest version of jsonschema though?

ramou:
My local testing works just fine (but possibly on jsonschema 2.6.), but based on some issues on other peoples guthubs that use Travis, the issue primarily seems to be related to Travis' recipe for Python 2.7 and 3.6. If you look at its reports for other python versions, you may note that pip freeze indicates the latter jsonschema 3.* is used and it runs fine in those builds.

I trimmed down the requirements.txt in that big PR prior to 0.5.19, and the jsonschema==2.6.0 was one of the few things that was left in. I'm confused as to why that was getting ignored during the Travis build (you'd think pip install . would have seen it in requirements.txt and done the downgrade, but maybe I don't get exactly what pips goal is... though I've read that it doesn't want to mind dependencies) and I don't know enough about what the differences are going to the new version to say we should upgrade our requirements.

I guess we just need to confirm that the new jsonschema verison (3.0.whatever) is ready for prime-time with boutiques, but that issue is separate from this one (where Travis is ignoring our version requirement for 2.6.0).

glatard:
ok, so let's merge this.

ramou:
Now I'm going to heck that all my other PRs point to develop :P

PR #447, opened 2019-03-16
Adjusted rounding system for exec simulate simulated floating point params

### LINKED ISSUE #329: Simulated values may violate exclusive min and max
This is one of the reasons why simulation tests still randomly fail. Let's add other ones here as we notice them.

### LINKED ISSUE #442: test_simulate fails randomly, but rarely when run.
I think related to #329 and possibly #337

It seems that random values outside of acceptable ranges can be generated.

ramou:
Closes #442 

May also close #329 

These changes create a variable `number_decimals` separate from `nd` in `localExec`'s `_randomFillInDict` function, that explicitly represents the number of decimal places to round to. `nd` is described as being the length of randomly generated strings. Given `number_decimals`, an `epsilon` is assigned acting as the upper bound error (traditional in this context) except we set it and use it to enforce exclusion as rounding may invalidate our boundaries given our current approach. Specifically, `random.uniform` is inclusive on both bounds, and if we want to exclusively generate something we are changing the specified bounds with our epsilon because we'll round it later.

While this change fixes the bug, it may be worth considering changes that more accurately reflect the exclusivity, such as checking that the resulting `random.uniform` is not at a boundary, and nudging it by a much finer `epsilon` if it is, skipping rounding altogether. This would allow descriptors to describe much finer minimums and maximums in much finer ranges than currently enforced by our `epsilon` of `0.001`. It would also render the distribution of simulated values more uniform.

Originally we had a hardcoded `epsilon` of `0.001` but rounded to two decimal places. On top of that, the affected input, `num_input` was not guaranteed to be used in the simulation. As such, this test only rarely failed.

These changes also add boundary tests on the min and max of floating point numbers being simulated, both inclusively and exclusively. The tests use the standard  `example1_docker.json` but force `num_inputs` to be required and removes the `groups` entry in that descriptor. mock is used to fix the random generation and the descriptor is manipulated before each use to get the appropriate checks.

Integer boundary tests should be considered, in particular, to verify that random numbers generated are properly being bound as expected. It may also want to be considered if the rounding keeps the distribution of potential values uniform or if we care about that.

### DIFF
--- a/tools/python/boutiques/localExec.py
+++ b/tools/python/boutiques/localExec.py
@@ -553,6 +553,13 @@ class LocalExecutor(object):
         # nd = number of random characters to use in generating strings,
         # nl = max number of random list items
         nd, nl = 2, 5
+        # number_decimals = the number of decimal places to round to
+        # on randomly generated floating point numbers
+        number_decimals = 3
+        # epsilon = an upper bound on the relative error due to
+        # rounding for the chosen number of decimals, making sure
+        # excluded values aren't rounded to by accident.
+        epsilon = 1.0/(10**number_decimals)
 
         def randDigs():
             # Generate random string of digits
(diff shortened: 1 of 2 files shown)


glatard:
Great work!

PR #457, opened 2019-03-29
Handle decoder errors when decoding stdout and stderr

### LINKED ISSUE #456: Non utf-8 encoded strings in stdout/stderr cause executor to crash
When running `bosh exec launch`, if the `stdout` or `stderr` of the program contains non-utf-8 encoded characters, the executor crashes with the error:
```
  File "/home/erin/PycharmProjects/boutiques/tools/python/env/boutiques/bin/bosh", line 11, in <module>
    load_entry_point('boutiques', 'console_scripts', 'bosh')()
  File "/home/erin/PycharmProjects/boutiques/tools/python/boutiques/bosh.py", line 526, in bosh
    out = execute(*params)
  File "/home/erin/PycharmProjects/boutiques/tools/python/boutiques/bosh.py", line 140, in execute
    return executor.execute(results.volumes)
  File "/home/erin/PycharmProjects/boutiques/tools/python/boutiques/localExec.py", line 386, in execute
    container_command, container_location)
  File "/home/erin/PycharmProjects/boutiques/tools/python/boutiques/localExec.py", line 33, in __init__
    self.stderr = stderr.decode("utf=8")
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x90 in position 1973: invalid start byte
```
Using Python 3.6 on Ubuntu.

erinb90:
Fixes #456 

`bosh exec launch` would crash with a `UnicodeDecodeError` if the stdout or stderr of a program contained non-decodable characters, such as letters with accents or other weird characters. I noticed it when testing a Boutiques descriptor exported from Nipype (fsl topup). 

To fix it, I set the error handling of the decoder not to crash when it encounters such characters. The handling differs depending on the Python version:
- Python 2.7: ignores the character entirely (still crashed when I tried character replacement)
- Python 3.4: replaces character with a �
- Python >=3.5: replaces character with an escape sequence (I find this is nicer than the �, but it was only introduced in 3.5)

Added a new example descriptor and example tool to test this.


glatard:
Great, thanks!

Issue #458, opened 2019-04-07
Input file lists are not supported


erinb90:
When one of the inputs is a list of files, `bosh exec launch` crashes with the following error:

```
File "/home/erin/PycharmProjects/boutiques/tools/python/env/boutiques/bin/bosh", line 11, in <module>
    load_entry_point('boutiques', 'console_scripts', 'bosh')()
  File "/home/erin/PycharmProjects/boutiques/tools/python/boutiques/bosh.py", line 534, in bosh
    out = execute(*params)
  File "/home/erin/PycharmProjects/boutiques/tools/python/boutiques/bosh.py", line 143, in execute
    results.skip_data_collection})
  File "/home/erin/PycharmProjects/boutiques/tools/python/boutiques/localExec.py", line 179, in __init__
    self.readInput(self.invocation)
  File "/home/erin/PycharmProjects/boutiques/tools/python/boutiques/localExec.py", line 837, in readInput
    self.public_in = self._generatePublicInvocation()
  File "/home/erin/PycharmProjects/boutiques/tools/python/boutiques/localExec.py", line 1226, in _generatePublicInvocation
    public_in_dict[id] = self._buildPublicFile(path)
  File "/home/erin/PycharmProjects/boutiques/tools/python/boutiques/localExec.py", line 1254, in _buildPublicFile
    filename = extractFileName(path)
  File "/home/erin/PycharmProjects/boutiques/tools/python/boutiques/localExec.py", line 1337, in extractFileName
    return os.path.basename(path)
  File "/home/erin/PycharmProjects/boutiques/tools/python/env/boutiques/lib/python3.6/posixpath.py", line 146, in basename
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not list
```
It seems that when the input is a file, the executor assumes it's a single file and does not handle the case when the input is a list.

PR #461, opened 2019-04-10
Support lists of input files

erinb90:
Fix for #458.

To test this, added a new input to `example1_docker.json` that takes a list of files. Published a new version of this file on Zenodo: https://zenodo.org/record/2634310.


glatard:
nice catch, great fix!

Issue #450, opened 2019-03-21
Bosh search doesn't consider all results when ordering by most downloaded


erinb90:
By default, `bosh search` will show the top 10 results ordered by "best match". These results will then be reordered by greatest number of downloads. However, this reordering will only be applied to the 10 results originally returned, not of all the results.

For example, `bosh search` currently returns this:
```
ID              TITLE                                   DESCRIPTION                                    DOWNLOADS
zenodo.1451003  qeeg                                    qeeg application                                      21
zenodo.2587157  blast_formatter                         Stand-alone BLAST formatter client, vers...           16
zenodo.2587156  blastdbcheck                            BLAST database integrity and validity ch...           14
zenodo.1450993  PostFreeSurferPipelineBatch-CentOS7     PostFreeSurferPipelineBatch HCP pipeline              10
zenodo.1484547  BIDS App - FreeSurfer 6.0               BIDS App version of freesurfer 6.0, from...           10
zenodo.1450997  FreeSurferPipelineBatch-CentOS7         FreeSurferPipelineBatch HCP pipeline                   9
zenodo.2566455  BIDS App - FSL Diffusion Preprocessing  Preprocessing pipeline for diffusion MRI...            6
zenodo.2566443  mask2boundary                           Transforms a binary mask of a niftii ima...            6
zenodo.2541125  BEst                                    <p>EEG/MEG source localisation technique...            4
zenodo.2566450  BIDS App - ndmg                         ndmg connectome estimation pipeline                    4

```
"Example Boutiques Tool", with 11,571 downloads, is not shown, because it wasn't part of the first 10 results before they were reordered. 

Unfortunately there does not seem to be a way to order by number of downloads when calling Zenodo API. Therefore I see a few possible options:
1. Perform the search in order by "most viewed" instead of "best match". In general, descriptors with a lot of downloads will also have a lot of views, but the results still won't be entirely accurate.
2. Once we know the total number of results, perform the search again, setting `--max` to the total. Reorder all the results by most downloaded, then only display the first 10. This is the most accurate but also the least efficient as it involves an additional API call.
3. Perform the search in order by "most viewed" and display the most viewed instead of the most downloaded. This is simple but changes the requirements of `bosh search`.

ramou:
[...]

PR #465, opened 2019-04-11
Changed bosh search to get and sort all results

erinb90:
#450 

Set the search to always return a maximum of 9999 results. Right now there are 33 descriptors and I can't imagine we'll ever have so many that this API call will become too slow (maybe one day :stuck_out_tongue:). It then sorts all the results by number of downloads and truncates the results list according to `max_results`.

### DIFF
--- a/tools/python/boutiques/searcher.py
+++ b/tools/python/boutiques/searcher.py
@@ -4,7 +4,6 @@ import requests
 import sys
 from collections import OrderedDict
 import numbers
-import json
 from operator import itemgetter
 from boutiques.logger import raise_error, print_info
 from boutiques.publisher import ZenodoError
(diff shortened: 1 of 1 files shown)


tools/python/boutiques/searcher.py
```diff
@@ -44,18 +43,21 @@ def __init__(self, query, verbose=False, sandbox=False, max_results=None,
...
             return self.create_results_list_verbose(results.json())
         return self.create_results_list(results.json())
 
     def zenodo_search(self):
+        # Get all results
         r = requests.get(self.zenodo_endpoint + '/api/records/?q=%s&'
                          'keywords=boutiques&keywords=schema&'
                          'keywords=version&file_type=json&type=software'
-                         '&page=1&size=%s' % (self.query, self.max_results))
+                         '&page=1&size=%s' % (self.query, 9999))
```
glatard: Would it make sense to define a variable for 9999 in the unlikely event that this value is updated in the future?

glatard:
Very useful! Looking forward to seeing the search results exceed memory size :)

PR #462, opened 2019-04-10
Add data record management

friesenA:
Issue #344 
This PR adds the user interface for data record management. Users are now able to 

- `bosh data inspect` : to view files in cache and an example
- `bosh data publish` : to publish datasets to zenodo individually or in bulk
- `bosh data delete` : to remove files from cache individually or clear the cache

Outstanding issues: Have been resolved.

Issue #385, opened 2019-01-16
Allow value-only tags


erinb90:
Right now tags must be in the form `key:value`. It would be good if we could also specify just values without a key. For example:
```
"tags": {
    "foo": "bar",
    "foo"
 }
```
Not sure if this is possible with the current version of JSON schema.

glatard:
It isn't possible indeed, and I agree it would be nice to have that as the need occurred in a few situations. Not sure if we should remove key-value tags, which would make descriptors non backward compatible, or just support both.

ramou:
I think we can get away with supporting both and with minimal effort being backwards compatible:

```
	"tags": [
		"blast", 
		"bioinformatics", 
		{
			"domain":[
				"makedatabse", 
				"foo", 
				"bar"
			]
		}
	]
```
which I would envision producing keywords: `blast`, `bioinformatics`, `domain:makeblastdb`, `domain:foo`, `domain:bar`

ramou:
Well, not one line:

```
foo = json.loads('{"tags": {"domain":["makedatabse", "foo", "bar"]}}')
bar = json.loads('{"tags": [{"domain":["makedatabse", "foo", "bar"]}]}')

foo == bar

if isinstance(foo["tags"], dict):
    newtags = []
    for key,val in foo["tags"].iteritems():
        newtags.append({key:val})
    foo["tags"] = newtags
	
foo == bar
```

ramou:
Python list comprehension let's me squeeze it a bit more
```
if isinstance(foo["tags"], dict):
    foo["tags"] = [{key:val} for key,val in foo["tags"].iteritems()]
```
Two lines. It can be squeezed to one except we'd cross 80 columns.

ramou:
The changes are straight-forward within boutiques, and I'm happy with the minimalist approach to nudging. However, when updating how we export to carmin, I will need some guidance, if @glatard can direct me. Specifically, we are currently assigning the existing key/value pairs to `carmin_desc['properties']` in line 78 of exporter.py. I'll need an idea of where to look to understand how they want their properties used and depending on that, an idea of how we would want to fudge our alternate approach.

ramou:
The best I've found so far is:
https://github.com/CARMIN-org/CARMIN-API/blob/23cb67bc6a641e2291b55fc1f58aa77ec09808e2/carmin.yaml#L584-L585

That last bit about being used in the `supportedPipelineProperties` concerns me. That said, it appears somewhat cyclic in its definition:
[...]

PR #466, opened 2019-04-12
Value-only tags

glatard:
#385 

The format initially suggested in the issue:
```
"tags": {
    "foo": "bar",
    "foo"
 }
```
isn't valid JSON: JSON properties have to be key-value pairs.

The other solution proposed in the issue:
```
"tags": [
            "neuroinformatics",
            "diffusion",
            "mri",
            "bids",
            {"foo": [
                "bar",
                "zap"
            ]}        
    ]
```
isn't backward compatible since the type of `tags` would be updated from `array` to `object`: all the existing descriptors would have to be updated, or the schema would have to support both syntaxes (which would be cumbersome).

Instead, this PR implements the following syntax:
```
"tags": {
    "foo": "bar",
    "foo1": true
 }
```

The only required modification in the schema is to accept values of type `boolean` in tags. The publisher is updated accordingly. No update is required in the CARMIN exporter.

N.B.: this PR includes commits from #462 to avoid future conflicts. Code should be reviewed only when #462 is merged.

@ramou @gkiar @erinb90


glatard:
This is now ready for review.

erinb90:
Looks good! However, there is a diff on `good.json` (from running local tests) that you might want to revert.

Issue #472, opened 2019-04-24
Add option to preserve quotes in generated command line


erinb90:
For example, Spark requires that its list arguments be surrounded by double quotes:
`"workspace-path/fmri1.mnc,workspace-path/fmri2.mnc" workspace-path/mask.mnc 100 "10 2 30" 20 0.001 preserveAtomsDict 0 dictionaryInitMethod DataElements blockWindowLength "10 1 30"`

To preserve those double quotes, I have to hardcode and escape them in the descriptor's command line, e.g. `\"[FMRI_DATA]\"`. This causes issues if it's an optional parameter and you don't put a value, because you will end up with `""` in the command line. It also causes issues if you have a command line flag, because it will end up looking like this: `"flag 1 2 3"`, when what you want is this: `flag "1 2 3"`.

It would be nice if the schema had an option to specify which character should surround an input.

PR #473, opened 2019-04-24
Add option to include double quotes around input value

erinb90:
#472 

Added an option `use-quotes` to place double quotes around an input value. Useful for cases like the SPARK EEG pipeline, which requires that its list inputs be surrounded by double quotes.


tools/python/boutiques/localExec.py
```diff
@@ -902,9 +903,12 @@ def escape_string(s):
...
                             s_val += s + list_sep
-                    val = s_val
-                elif escape:
-                    val = escape_string(val)
+                    val = s_val if surr_char is None else s_val + surr_char
+                else:
+                    if escape:
+                        val = escape_string(val)
+                    if surr_char is not None:
+                        val = surr_char + str(in_out_dict[param_id]) + surr_char
```
glatard: This breaks the escaping of special characters in case `surr_char` isn't the double quote character. 
`val` should always be escaped after `surr_char` was appended. This is also a problem with lists. For instance, the command line generated from https://gist.github.com/glatard/a9d1fafb35d21a4fcbf111eedd5c0b60 should be `echo '-file name with space.tex-' '-filename with a ;.pdf-'`, and not `echo -file name with space.tex- -filename with a ;.pdf-`

erinb90: Ok, so would it be sufficient to just add `val = escape_string(val)` after adding the surrounding characters?

glatard: Yes I think so, but only `if escape`. For instance, numbers shouldn't be escaped.

glatard:
Hi @erinb90, I don't think this works as it still adds the surrounding characters _after_ it escaped the parameter value. This will break if the surrounding character is a special character, for instance ';'.

erinb90:
> Hi @erinb90, I don't think this works as it still adds the surrounding characters _after_ it escaped the parameter value. This will break if the surrounding character is a special character, for instance ';'.

Oh so you mean the single quotes should be outside the surrounding character? That would make much more sense. I was was just simulating command lines without actually trying to run them in a terminal :)

erinb90:
I'm wondering if there would ever be a use case for a surrounding character that isn't the double quote? I've never seen a command line input surrounded by ; or % or anything like that. Maybe instead of specifying the surrounding character, we can instead just have a flag called 'keep-quotes' that would just surround an input with double quotes?

glatard:
Indeed that might be simpler. In that case you should still escape the string in case it contains a double quote, but I'm not sure how `shlex.quote` could do that. The best would be to replace `"` with `\"` yourself in the original string.

glatard:
Hi @erinb90, sorry but this still breaks when the input contains a '\"'... For instance, the following invocation:
```
{
  "param_str": "fo\"o",
  "param_list": ["string with a ; in it", "string2"],
  "param_file": "file name with space.tex",
  "param_num": 1
}
```
results in the following command:
```
$ bosh exec simulate tools/python/boutiques/schema/examples/use_quotes.json -i tools/python/boutiques/schema/examples/use_quotes_inv.json
Generated Command:
-s "'fo"o'" "'string with a ; in it' string2" "'file name with space.tex'" "1"
```
Note the presence of an extra double quote inside string `foo`. I think you should check for the presence of double quotes in the input strings, and replace them with `\"` or even `\\\"`.

Escape strings are such a nightmare!

erinb90:
Ok, I have something now that I think should work. With invocation:
```
{
  "param_str": "fo\"o",
  "param_list": ["string with a ; in it", "string2"],
  "param_file": "file name with space.tex",
  "param_num": 1
}
```
we get command line: 
`echo -s "'f\"oo'" "'string with a ; in it' string2" "'file name with space.tex'" "1"`

and the program outputs: 
`-s 'f"oo' 'string with a ; in it' string2 'file name with space.tex' 1`

glatard:
Very sorry for not seeing that before but I think there's another issue here. In your example, `"'file name with space.tex'"` is not correct because it would refer to a file name with single quotes in it. It should be `"file name with space.tex"` instead. So we should not add the single quotes when double quotes are added. 

See the issue here:
```
# Create a file with a space in its name
$ touch a\ b.txt

# File can be access through single quotes only:
$ file 'a b.txt'
a b.txt: empty

# File isn't found when double quotes are added:
$ file "'a b.txt'"
'a b.txt': cannot open `'a b.txt''
```
At the same time, removing the single-quote escape wouldn't work if the file name contains a `$`, so I'm not sure how this should be handled for now.

glatard:
This seems to be a rabbit hole, might create more issues than it solves, and is only specific to one particular pipeline for now. As discussed with @erinb90, we will close this and get back to it if needed.

Issue #390, opened 2019-01-22
Add descriptor2function utility


gkiar:
This utility would take a descriptor (by any of the normal acceptable means), validate it, and return a function with the descriptor inputs as arguments which returns the output fields.

This would make it really easy to consume multiple Boutiques descriptors in a single Python workflow.

glatard:
It would also provide a Python API to non-Python tools.

PR #474, opened 2019-04-30
Descriptor to function

glatard:
#390 

This creates a function to execute or simulate the execution of a tool from its descriptor:

Execute:
```
from boutiques.descriptor2func import function
mcflirt = function('zenodo.2602109')  
mcflirt(in_file='/home/glatard/data/test.nii.gz') # default behaviour is launch
```

Execute, with options:
```
from boutiques.descriptor2func import function
mcflirt = function('zenodo.2602109')  
mcflirt('-s', in_file='/home/glatard/data/test.nii.gz') # default behaviour is launch
```

Simulate:
```
from boutiques.descriptor2func import function
mcflirt = function('zenodo.2602109')
mcflirt('simulate', in_file='/home/glatard/data/test.nii.gz')
```


glatard:
Hey @gkiar, would you have a minute to look at that?

tools/python/boutiques/descriptor2func.py
```diff
@@ -0,0 +1,27 @@
...
+from boutiques import validate, execute, prettyprint
+import json
+
+
+def function(descriptor, name=None, mode='launch'):
+
+    validate(descriptor)
+    descriptor_json = loadJson(descriptor)
+
+    def f(*args, **kwargs):
```
gkiar: Can we document that `args` is used for `bosh exec launch` runtime arguments, and invocations are passed in via `kwargs`? Just to make sure this is obvious to people.

tools/python/boutiques/descriptor2func.py
```diff
@@ -0,0 +1,27 @@
...
+        if mode == 'launch':
```
gkiar: I'd like for the function to be generic enough for us to use either mode without redefining it... any ideas how we can pass this option on the fly?

tools/python/boutiques/descriptor2func.py
```diff
@@ -0,0 +1,27 @@
...
+            return execute(mode, descriptor,
+                           '-i', json.dumps(kwargs), *args)
+        else:
+            raise Exception("Unsupported mode: {}".format(mode))
+
+    if name is None:
+        f.__name__ = str(descriptor_json['name'])
+    else:
+        f.__name__ = name
+    f.__doc__ = prettyprint(descriptor)
```
gkiar: Since this is a lot of text, might be worth pruning down to the last bit that is the input params? We could just grep the heading for them and do after that. I don't have a strong feeling here, but that may make the helptext more easily parseable, esp for big descriptors.

tools/python/boutiques/tests/test_desc2func.py
```diff
@@ -0,0 +1,24 @@
...
+from boutiques.descriptor2func import function
+if sys.version_info < (2, 7):
+    from unittest2 import TestCase
+else:
+    from unittest import TestCase
+
+
+class TestDesc2func(TestCase):
+
+    def test_desc2func(self):
```
gkiar: can we also test the exec mode? Just a sanity check that things are being passed properly

tools/python/boutiques/descriptor2func.py
gkiar: Maybe this would be done by creating the `__doc__` with details about this function wrapper, and appending the `prettyprint` to it for actual tool docs

tools/python/boutiques/tests/test_desc2func.py
glatard: done

tools/python/boutiques/descriptor2func.py
glatard: done

tools/python/boutiques/descriptor2func.py
glatard: done

tools/python/boutiques/descriptor2func.py
glatard: yeah, I wouldn't like to have to post-process pprint here. Maybe we could add various options to pprint to show the different sections.

gkiar:
Looks great, thanks!!!

Issue #254, opened 2018-04-20
Document Python API


glatard:
Including exceptions thrown, returned values, etc.

glatard:
For instance, `help(bosh.execute)` should return something...

PR #479, opened 2019-05-08
GitHub Templates for issue and pull request

mathdugre:
Created a set of template to make the issues and pull request more uniform.

### Purpose
This should help in making issues (pull request) more comprehensible thus improving the reviewing process. Also this aims at improving the code base quality by establishing standard and make sure that their adoption is made.

## Minor suggestions
- Document the API according to [PEP 257](https://www.python.org/dev/peps/pep-0257/)

## Major suggestions
- Making all code compliant with the [PEP 8](https://www.python.org/dev/peps/pep-0008/) code style
- Adopting [sphinx](http://www.sphinx-doc.org/en/master/) as a documentation lint
- Using type annotation with [mypy](http://www.mypy-lang.org/) as a linter


glatard:
Hi @mathdugre, thanks for doing this!

The templates look great to me! Just out of curiosity, did you take them from somewhere in particular?

I definitely agree with documenting the API and putting it on sphinx. This should be part of #254.

As for code style, the code should already be compliant with PEP8, and we test that using `pycodestyle` in Travis. Do you think we still need to make it a check box in the PR template?

As for `mypy`, I don't know. I'm not sure that we want to do this at all in Boutiques, it's quite a major change... Maybe we can have this conversation in a separate issue so that it doesn't hinder the merging of this one.

@gkiar @ramou

mathdugre:
Hi @glatard for the issue templates I was strongly inspired by [this one](https://github.com/stevemao/github-issue-templates/blob/master/bugs-only/ISSUE_TEMPLATE.md). For the PR templates I got inspiration from a few ones and cherry picked the features I liked from them.

I will start working on the documentation.

For pep8, I pointed it out since I came across a few function CapWord convention and unsorted import in some modules. Also some tool like [black](https://github.com/python/black) could ease the process of formating the code to comply with PEP8. I'll open another issue for discussion.

As for mypy I totally agree that it would be a major change. I'll open an issue to discuss the consequences that it would imply.

For now I'll remove PEP8 and mypy from the checklist.
Also, do you think a discussion template could be useful?

glatard:
As for PEP8, pycodestyle could be made more stringent in Travis. Black also looks relevant, however, we should first update all the code before requesting that new PRs do that.

The same goes with API documentation and Sphinx: before we have them in the template, we should make sure that the current code complies to them, otherwise it will be difficult for developers to adopt a standard which is currently not enforced.

For bug reports, I'm worried that the template is a bit verbose and will discourage people reporting bugs. Maybe we can indicate that some fields are optional, in their title?

erinb90:
Hey @mathdugre, this is really nice! If possible, I would also add templates for refactoring existing code and for creating/updating documentation.

glatard:
Let's merge this as it's already super useful! I agree with @erinb90, we could have templates to refactor code and documentation. Also, the bug PR template should include tests too.

mathdugre:
I will work on the refactoring and documentation template. 


> Also, the bug PR template should include tests too.

@glatard I am not sure to understand what is meant here. Is it the check box with label `DO Created unit test for the feature.` ?

glatard:
yes, this it what I meant.

PR #485, opened 2019-05-24
Json crash on bosh validate#481

DarrinFong:
Changed from json.loads to simplejson.loads for more verbosity (also changed the code outside of the issue's description while we're there)

```
bosh validate invalid_json.json

[...] JSONDecodeError("Expecting ',' delimiter or '}'", s, end - 1)
[...] Expecting ',' delimiter or '}': line 9 column 2 (char 243)
```


glatard:
Hi @DarrinFong, this looks great! Please check why the tests are failing and we could merge that.

tools/python/boutiques/publisher.py
```diff
@@ -4,6 +4,7 @@
 from boutiques.logger import raise_error, print_info
 from boutiques.zenodoHelper import ZenodoError, ZenodoHelper
 import json
```
glatard: Can we remove `import json`?

tools/python/boutiques/tests/test_simulate.py
```diff
@@ -3,6 +3,7 @@
 import os
 from unittest import TestCase
 import json
```
glatard: Same comment.

tools/python/boutiques/util/utils.py
```diff
@@ -1,5 +1,6 @@
 import os
 import json
```
glatard: Same comment.

DarrinFong:
Reviewing this to effectively remove all usage of json and replace with simplejson

gkiar:
@glatard when merging this, because there was a singularity image added and removed, we should make sure to merge with a squash-commit which will just take the final changes and leave anything transient, rather than bring its bulk into history.

tools/python/boutiques/publisher.py
DarrinFong: Replaces all usage of json package with simplejson

DarrinFong:
Accidentally rebased commits that fix docker/singularity issue on Travis

DarrinFong:
This is messy, I'll make a cleaner branch and PR that instead

Issue #373, opened 2019-01-09
`bosh example` should produce invocations with no optional parameter


glatard:
This would make it easier for users to start from a minimal invocation when tool has a lot of optional parameters. A flag, for instance `-c` or `--complete` would trigger complete invocations, including optional parameters.

DarrinFong:
I don't quite understand the term invocation in this issue (or inside of boutiques)

After this issue is implemented, does it mean the JSON output of __this__ command will contain the most basic parameters?

```
bosh example good.json
{
    "non_optional1": "foo",
    "non_optional2: "bar"
}
```
and with the ```--complete``` flag, the result would be
```
bosh example -c good.json
{
    "non_optional1": "foo",
    "non_optional2: "bar",
    "optional1: "foo",
    "optional2: "bar"
}
```

glatard:
yes exactly! Assuming that "most basic parameters" means "non-optional parameters".

DarrinFong:
How do I determine which param in the mutex group to take? 
For example which one between these two in good.json
```
{
    "dti_file": "f_dti_file_65.nii.gz",
    "dti_file_minc": "f_dti_file_minc_85.j"
}
```

glatard:
I would just pick one randomly.

PR #494, opened 2019-05-31
Invocations without optional params #373

DarrinFong:
```bosh example [descriptor].json``` will output sample invocation with random optional params depending on predefined groups (ie: mutex, all-or-none, one-is-required). 

Adding the ```--complete``` flag will output sample invocation with all optional params (still depending on groups).

Can't exclusively display required fields when the ```--complete``` flag is missing without changing localExec because of the groups.

### CODE REVIEW

**Comment by glatard** [comment]
Hi @DarrinFong, the code looks fine, but I'm not sure that it's doing what it's supposed to. For instance, the following call returns a lot of optional parameters, such as `bins`, `cost` and others.

```
$ bosh example zenodo.2602109
{
    "bins": -48,
    "cost": "woods",
    "dof": 5,
    "in_file": "f_in_file_58.mnc",
    "init": "f_init_92.j",
    "interpolation": "spline_final",
    "mean_vol": false,
    "rotation": -19,
    "save_mats": false,
    "save_plots": true,
    "save_rms": true,
    "scaling": 19.163,
    "smooth": -11.201,
    "stages": -22,
    "stats_imgs": true,
    "use_contour": false,
    "use_gradient": true
}
```
This is misleading as it gives the impression that a minimal working example needs all these parameters to be filled in while only `in_file` is required. I expect this example to return only ```{"in_file": 'some_file.nii.gz"}```. Does it make sense?

**Comment by DarrinFong** [comment]
It was because of the code beginning at:
```
# Choose a random number of times to try to fill optional inputs
   opts = [p for p in self.input if self.safeGet(p['id'], '') in [None, True]]
```
I don't quite understand why it needs to insert random parameters but I committed changes and the results should now be correct!

File: tools/python/boutiques/tests/test_example.py [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+                                                 'example-invocation/'
+                                                 'example_descriptor.json')
+        command = ("bosh example " + descriptor)
+        process = subprocess.Popen(command, shell=True,
+                                   stdout=subprocess.PIPE)
+        output = json.loads(process.stdout.read())
+
+        # Can't assert more than the required params because
+        # params are randomly selected
+        self.assertDictContainsSubset({"b1": "b1", "c2": "c2"}, output)
```
glatard: I think we'd want equality of the two dicts here, not subset. Do you agree? Also, `assertDictContainsSubset` seems to be deprecated since 3.2, do you think you could use something else?

DarrinFong: Indeed, ```example``` yielded flaky results before my last commit and I forgot to change the tests. Using ```assertDictEqual``` now.

glatard:
alright, time to merge this, thanks for all the updates!

Issue #486, opened 2019-05-28
BIDS Execution property


glatard:
Add a BIDSExecVersion field to the schema, to specify that a descriptor describes a BIDSApp. See discussion in https://docs.google.com/document/d/104HLZedFtx0TaXEUwd7eyWvJUlc0CcSUtCzwjNgmGxE/edit?ts=5ced5e9f. 

We could either:
(1) add this field to the schema root
(2) use a free-form tag (key-value pair) for it
(3) use a custom property (can be any object)
I'm leaning toward (1) or (3) given that (2) wouldn't technically be part of the schema and would therefore be prone to typos. I guess (1) makes sense. 

What do people think?

glatard:
(3) is also not right, since that property wouldn't be part of the schema. So (1) seems to be the way to go!

jbpoline:
But BIDSExecVersion would be only required for a BIDS app ? so the boutique validator would have a --boutique argument to check that BIDSExecVersion exists only in the case of a BIDS app ?

glatard:
Yes, something like that.

PR #496, opened 2019-06-03
(WIP) BIDS execution version

glatard:
#486 

This adds a `bids-exec-version` property to the schema, which can be used by BIDS apps to specify a given version of the BIDS App execution specification. The property is a free-form string but could be changed to an enum or regexp if needed. The property was added to `bids_good.json` for testing purpose, although the tool doesn't comply to BIDSexecution yet.

@gkiar @effigies

### DIFF
--- a/tools/python/boutiques/schema/descriptor.schema.json
+++ b/tools/python/boutiques/schema/descriptor.schema.json
@@ -16,6 +16,12 @@
             "description": "Tool version.",
             "type": "string"
         },
+        "bids-exec-version": {
+            "id": "http://github.com/boutiques/boutiques-schema/bids-exec-version",
+            "minLength": 1,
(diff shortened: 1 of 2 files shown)


effigies:
Seems reasonable, but I would suggest waiting until we have a full draft to merge, in case people have concerns about this field (or suggest an alternative name).

glatard:
We should update this so that we can easily add other derived specs without having to updated the schema root, for instance using an enum.

gkiar:
Note: the updated execution spec proposal is [here](https://docs.google.com/document/d/104HLZedFtx0TaXEUwd7eyWvJUlc0CcSUtCzwjNgmGxE/edit?disco=AAAAEG9-PkM&ts=5e1de576&usp_dm=true)

PR #513, opened 2019-06-10
Bexample omit false values#504

DarrinFong:
## Previous Behavior
"false" is occasionally generated for inputs of type "Flag", resulting in invalid invocations.

## Changed Behavior
Values generated for inputs of type "Flag" will always be "true".
"false" values are those that are omitted
The internal descriptor validator in localexec will check for params of type "flag" and throw an error if the value is set to false

Can run ```bosh example zenodo.3240521 -c``` and none of the values should be false


glatard:
Thinking about that, I think that all flags should be made optional, meaning that:
(1) `bosh example` shouldn't use any flag in the returned invocation.
(2) `optional: false` shouldn't be valid for an input of type "Flag": we should specify and document that in the schema.

Technically, this breaks backward compatibility but I don't think that we have any descriptor with mandatory flags published yet. So it's more a loophole in the current schema than an enhancement.

What do you think @gkiar and @ramou? Hopefully we all agree and @DarrinFong could add this to this PR, as it is very much related. Otherwise we will handle the discussion in a separate issue.

ramou:
I think the second point, in particular is relevant. I could see
required:one-of at the group level, but otherwise flags are implicitly
optional and capturing that semantic makes sense to me.

glatard:
@gkiar, what do you think?

DarrinFong:
@glatard Changes were made!
(1) ```bosh example``` doesn't return invocation with flags unless ```-c``` flag is included.
(2) validator already checks for non optional flags in inputs

glatard:
Now I'm back to getting optional parameters in the example:
```
$ bosh example zenodo.3240521
{
    "h": -13,
    "input_file": "f_input_file_65.cpp"
}
```
Perhaps `test_example.py` should be improved to check that too.

DarrinFong:
@glatard I'm not 100% sure but I think ```input_file``` is in a group that requires another input. 
Ex: running ```bosh invocation zenodo.3240521 -i '{"input_file": "f_input_file_65.cpp"}``` would fail due to missing required param.

glatard:
You're completely right! My bad, I hadn't realized that the descriptor was so smart :) Let's merge this, well done!

PR #514, opened 2019-06-11
Fix singularity binding arguments for mounting multiple volumes

jerdra:
## Changes

For an example call:

<code> bosh exec launch -v p1:/b1 -v p2:/b2 </code>

The following incorrect singularity call is generated:

<code> singularity exec --cleanenv -B p1:/b1 p2:/b2 </code>

This PR updates it such that the correct singularity call is produced:

<code> singularity exec --cleanenv -B p1:/b1 -B p2:/b2 </code> 

Let me know if there's anything else I should add (i.e tests for change - wasn't sure if it was needed because of the rather small change)


glatard:
Hi @jerdra, great catch and thank you so much for the fix! Indeed, this is important enough that we should test it. Adding two mount strings to the tests in https://github.com/boutiques/boutiques/blob/831f937a6b1491af63a800786967e4d9bca1e262/tools/python/boutiques/tests/test_example1.py#L231 would be enough I think. You could also test that the directories are actually mounted by testing their existence in https://github.com/boutiques/boutiques/blob/master/tools/python/boutiques/schema/examples/example1/exampleTool1.py

jerdra:
HI @glatard, will definitely work on implementing the test. However getting the tests up and running is proving to be a bit of a hassle. I've run into two issues:

### Issue 1
When running test_example1_exec_singularity which tries to use the descriptor:
schema/examples/example1/example1_sing.json

I get an error from singularity pull. Singularity replaces the docker:// index with shub:// resulting in the following error:

<code>
WARNING: pull for Docker Hub is not guaranteed to produce the
WARNING: same image on repeated pull. Use Singularity Registry
WARNING: (shub://) to pull exactly equivalent images.
Docker image path: index.docker.io/boutiques/example1:test
ERROR MANIFEST_UNKNOWN: manifest unknown
Cleaning up...
</code>

Upon checking DockerHub it seems that example1:test isn't available, using example1:latest produces the required singularity image. 

### Issue 2
What context are the tests run in a local environment? When running py.test in the tests folder the singularity test cannot find /setup.py in the container directory but there is no where in the Dockerfile in the example1 folder where setup.py is included. I'm assuming its coming from binding a local directory? Is it running the standard boutiques setup? 

Finally, is there any documentation regarding setup for running tests properly? That would be incredibly helpful!

jerdra:
Hi @glatard, I've made and validated the singularity mounting as asked for. Some other aspects of the testing setup also had to be changed as indicated in the latest commit.

Unfortunately I could not check other tests after making minor changes to example1 descriptor/invocation jsons since they rely on docker pulls for which i need sudo permissions - my lab isn't currently implementing the docker group.

jerdra:
So it seems like a bunch of tests were testing bosh's behaviour incorrectly (expecting crashes when using the valid descriptor json for example). I've amended these issues and was able to confirm that all tests that previously failed the CI check work.

Additionally, the tests:
<code>test_exec_singularity</code> and <code>test_exec_singularity_force_docker</code> both work locally but don't seem to work on Travis CI. 

Any thoughts on why this might be?

tools/python/boutiques/schema/examples/example1/example1_sing.json
```diff
@@ -1,7 +1,7 @@
 {
-    "command-line": "exampleTool1.py [CONFIG_FILE] [STRING_INPUT] [FLAG_INPUT] [NUMBER_INPUT] [ENUM_INPUT] [FILE_INPUT] [LOG] [INT_LIST_INPUT] [NO_OPTS]", 
+    "command-line": "exampleTool1.py [CONFIG_FILE] [STRING_INPUT] [FILE_LIST] [FLAG_INPUT] [NUMBER_INPUT] [ENUM_INPUT] [FILE_INPUT] [LOG] [INT_LIST_INPUT] [NO_OPTS]", 
     "container-image": {
-        "image": "boutiques/example1:test", 
+        "image": "boutiques/example1:latest",
```
glatard: `test` is the correct tag, it is the one built in `.travis.yml`.

tools/python/boutiques/schema/examples/example1/example1_sing.json
```diff
@@ -1,7 +1,7 @@
 {
-    "command-line": "exampleTool1.py [CONFIG_FILE] [STRING_INPUT] [FLAG_INPUT] [NUMBER_INPUT] [ENUM_INPUT] [FILE_INPUT] [LOG] [INT_LIST_INPUT] [NO_OPTS]", 
+    "command-line": "exampleTool1.py [CONFIG_FILE] [STRING_INPUT] [FILE_LIST] [FLAG_INPUT] [NUMBER_INPUT] [ENUM_INPUT] [FILE_INPUT] [LOG] [INT_LIST_INPUT] [NO_OPTS]",
```
glatard: Let's call this `MOUNT_LIST` for clarity.

tools/python/boutiques/tests/test_prepare.py
```diff
@@ -65,7 +65,7 @@ def test_prepare_sing(self):
         ret = bosh.execute("prepare",
                            os.path.join(example1_dir,
                                         "example1_sing.json"))
-        self.assertIn("Local (boutiques-example1-test.simg)", ret.stdout)
+        self.assertIn("Local (boutiques-example1-latest.simg)", ret.stdout)
```
glatard: should be `test` here too.

tools/python/boutiques/tests/test_example1.py
```diff
@@ -243,7 +245,9 @@ def test_example1_exec_singularity(self):
             bosh.execute("launch",
                          self.get_file_path("example1_sing.json"),
                          "-x",
-                         self.get_file_path("invocation_sing.json")),
+                         self.get_file_path("invocation_sing.json"),
+                         "-v","{}:/test_mount1".format(self.get_file_path("example1_sing_mount1")),
+                         "-v","{}:/test_mount2".format(self.get_file_path("example1_sing_mount2"))),
```
glatard: We should also test that the files get mounted in the container.

tools/python/boutiques/tests/test_prepare.py
```diff
@@ -101,10 +101,10 @@ def test_prepare_sing_image_does_not_exist(self):
         with pytest.raises(ExecutorError) as e:
             bosh.execute("prepare",
                          os.path.join(example1_dir,
-                                      "example1_sing.json"),
+                                      "example1_sing_crash_pull.json"),
```
glatard: Good point! The test was passing before due to `test` though. At some point we should mock that.

tools/python/boutiques/tests/test_example1.py
glatard: OK, this is already done :)

jerdra:
thanks for the feedback. I'll revert the changes on the container being pulled (latest --> test), as that was made under the (incorrect) assumption the local test environment mapped to Travis but that doesn't seem to be the case - apologies this is my first experience with travis.

glatard:
No worries at all! I'll send you a PR in a few minutes with a few fixes.

tools/python/boutiques/tests/test_prepare.py
jerdra: to clarify should we leave it as _sing_crash_pull or revert to _sing? 
I figured that giving an invalid container would be scenario that this test seeks to evalute, let me know if this is right

glatard:
OK, now the tests are failing because `test_example1_exec_docker_Index_force_singularity` needs the mount strings. This test was just merged in `develop`, do you think you could pull `develop` and add the mount string? We're getting there!

jerdra:
sure can do!

jerdra:
@glatard I did a bit of investigating for why this test is failing and I think I've figured out why:

It's because in the descriptor json for the typical singularity test there is the container option "--home $PWD:PWD". Therefore the singularity container starts in tools/python and is able to find ./subdir1/subdir2.

When using the example1_docker_w_index.json, the container option is "-e HOME=$PWD", which runs into an incompatible option for container warning resulting in no additional container option of remapping HOME in the container. Therefore the singularity container starts in the user's default $HOME and therefore cannot find ./subdir1/subdir2/config.txt.

A solution would be to map docker home environment setting to the equivalent singularity version when using --force-singularity. Let me know what you think!

glatard:
Hi @jerdra, this is in fact because some tests still used `example_sing_mount1` while this is now `example_mount1` as it's used in both Singularity and Docker. There was also issues with pycodestyle. My PR to your branch should fix it all. Hopefully we can merge this now!

jerdra:
thanks @glatard. Seems like my local instance and the Travis environment aren't aligned and so my local debugging findings don't seem to carry through. 

Any tips on how to mirror the Travis environment for testing? The test still fails in my local test environment for the reasons I mention above (unrelated to the actual mounting but rather home binding). 

Thanks again for your patience!

glatard:
Yeah, the testing environment can be a bit subtle. The [getting started guide](https://github.com/boutiques/boutiques/blob/master/GETTING_STARTED.md) might help but I trust you've been through it already.  The pre-install steps in https://github.com/boutiques/boutiques/blob/master/.travis.yml might help too. In particular, you will see that container image `boutiques/exemple1:test` is first built with Docker, then converted to Singularity and moved to the current working directory. This is the way we found to not have to rely on DockerHub during tests, as it's good practice for tests not to depend on external services. The version of Singularity is also a critical factor, as behaviour tends to change fast between versions. We have decided to test on the version available on Compute Canada, currently 2.6.1, but there might be errors in more recent versions. I'm currently using 3.2.1 on my laptop though, and the tests pass.

That said, the dropping of container options when forcing Singularity will be an issue, although I'm not sure why it doesn't show up in Travis. If you want, you could submit an issue on that showing a brief example where it fails.

In the meantime, let's merge this PR! Thanks for reporting and fixing it!

Issue #509, opened 2019-06-07
Running Docker images in Singularity transparently


glatard:
As per #298, we're supposed to run Docker images in Singularity when Docker is not available or when `--force-singularity` is used. This currently doesn't work when Docker images are pulled from DockerHub, which is the main use case. I think that the test in https://github.com/boutiques/boutiques/blob/master/tools/python/boutiques/tests/test_example1.py only works because the tool image is built locally and hence doesn't need to be pulled.

## Expected Behavior

When both Singularity and Docker are installed, the following command lines should all work successfully, assuming that `test.nii.gz` is present in the cwd (see example file [here](https://github.com/boutiques/tutorial/tree/master/notebooks/data)):
```
bosh exec launch zenodo.1482743 '{"infile": "./test.nii.gz", "maskfile": "test_brain.nii.gz"}'
```
```
bosh exec launch --force-singularity zenodo.1482743 '{"infile": "./test.nii.gz", "maskfile": "test_brain.nii.gz"}'
```
## Current Behavior

The command line with `--force-singularity` crashes as follows:
```
h exec launch --force-singularity zenodo.1482743 '{"infile": "./test.nii.gz", "maskfile": "test_brain.nii.gz"}'
docker is /usr/bin/docker
[ ERROR ] Could not pull Singularity image: 
 * Pull command: singularity pull --name "mcin-docker-fsl-latest.simg.tmp" index.docker.io://mcin/docker-fsl:latest
 * Error: FATAL:   Unsupported transport type: index.docker.io
```

This is primarily because `container-image/index` is set to `index.docker.io`. Removing the index entry leads to another error:
```
$ bosh exec launch --force-singularity ~/.cache/boutiques/zenodo-1482743.json '{"infile": "./test.nii.gz", "maskfile": "test_brain.nii.gz"}'
docker is /usr/bin/docker
[ ERROR ] Could not pull Singularity image: 
 * Pull command: singularity pull --name "mcin-docker-fsl-latest.simg.tmp" shub://mcin/docker-fsl:latest
 * Error: FATAL:   Failed to get manifest from Shub: The requested manifest was not found in singularity hub
```
This is because the index defaults to `shub://` when Singularity is used while it should be `docker://` in this case. Setting `container-image/index` to `docker://` solves the issue but it makes the execution break with Docker as `docker://` isn't a valid Docker index (it's a convention adopted by Singularity to specify that an image is stored on DockerHub).

## Possible Solution

[...]

PR #519, opened 2019-06-14
Docker in singularity transparently#509

DarrinFong:
When using ```--force-singularity``` flag
### Previous conIndex
```shub://[...]/image.sing```

### After changes
```docker://[...]/image.sing``` (if it's a docker image)


glatard:
Thanks for this @DarrinFong! In the future, could you refer to the related issue, here #509, in the PR body rather than in the title, so that the reference shows in the issue? The PR template should have a section for that.

Issue #524, opened 2019-07-15
"requires-inputs" field in descriptors should support groups


DarrinFong:
## Purpose

The "requires-inputs" field in descriptors currently only support inputs. Support for groups should be added to establish more complex dependencies.

#### Related issues

This is required to complete the implementation for: issue #392 

## Context

Currently it's impossible to have this command-line structure reinforced by the descriptor:
```
foo X a
foo X b
```
We can currently ensure:
```a``` -> ```X``` and ```b``` -> ```X```, but not ```X``` -> ```a``` ⊕ ```b```

PR #525, opened 2019-07-15
Docopt export descriptor#392

DarrinFong:
# Change
 _This is a 'partial' implementation and requires #524 to complete_

```bosh import dcpt [OUTPUT] [DOCOPT_PY_INPUT]```

Creates a valid boutiques descriptor from a valid docopt script inside of a .py file.

### CODE REVIEW

File: tools/python/boutiques/templates/template_docopt_descriptor.json [hunk 1]
```diff
@@ -0,0 +1,40 @@
+{
```
gkiar: As discussed, potentially use `bosh create` with no arguments to generate a template from which you can build?

tools/python/boutiques/schema/examples/docopt_to_argparse/sample_pattern.json
```diff
@@ -0,0 +1,38 @@
+{
```
gkiar: Unless we use these in tests (to verify the "expected" interpretation of command args), we should remove the `pattern` files from the repo

DarrinFong:
Flaky tests
https://travis-ci.org/DarrinFong/boutiques/builds/568490831

DarrinFong:
> I'm happy to merge - is it ready, or are you building any more in?

Cool! It's ready

gkiar:
Great! Nicely done :)

Issue #523, opened 2019-07-10
bosh publish --replace puts an invalid DOI in the descriptor


gkiar:
When typing `bosh publish --update` I'm able to successfully update my descriptor, but the `doi` field now has the full doi, i.e. `"doi": "10.5281/zenodo.3308620"` for the tool `oneVoxel`. This makes the local copy of the descriptor un-runnable, because when a DOI is found in the descriptor, `bosh` tries to open it, but only the Zenodo portion is used for this search so the descriptor cannot can't be found and the tool crashes. When I strip out the leading portion of the DOI manually, things work as expected.

## Possible Solution (optional)
Strip out this portion of the DOIs prior to adding them to the descriptor

PR #526, opened 2019-07-16
DOI patch for publisher

gkiar:
Currently, the publisher returns a full DOI in the `doi` field of the descriptor, including the characters preceding zenodo (i.e. `#######/zenodo.######`), but this makes descriptors invalid. I've added a safe stripping of this component preceding the `/` so publishing won't result in creating invalid descriptors anymore (😢)

### DIFF
--- a/tools/python/boutiques/localExec.py
+++ b/tools/python/boutiques/localExec.py
@@ -1225,7 +1225,12 @@ class LocalExecutor(object):
             # Most recent DOI in file if user is publisher
             # Include check to ensure descriptor is unmodified
             if self.desc_dict.get('doi') is not None:
-                doi = self.desc_dict.get('doi')
+                # Zenodo returns the full DOI, but for the purposes of
+                # Boutiques we just use the Zenodo-specific portion (as its the
+                # unique part). If the API updates on Zenodo to no longer
+                # provide the full DOI, this still works because it just grabs
+                # the last thing after the split.
+                doi = self.desc_dict.get('doi').split('/')[-1]
                 if loadJson(doi) == self.desc_dict:
                     return doi
             # DOI in filename if descriptor pulled from Zenodo
--- a/tools/python/boutiques/localExec.py
+++ b/tools/python/boutiques/localExec.py
@@ -1225,12 +1225,7 @@ class LocalExecutor(object):
             # Most recent DOI in file if user is publisher
             # Include check to ensure descriptor is unmodified
             if self.desc_dict.get('doi') is not None:
-                # Zenodo returns the full DOI, but for the purposes of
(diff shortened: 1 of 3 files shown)


glatard:
Not sure I understand the PR and #523 well. It seems to me that the problem is bosh trying to use the DOI as a Zenodo id, which it isn't. It makes sense to me that the DOI remains a valid doi, with its prefix, and that bosh strips the prefix when it needs a Zenodo id. Am I missing something?

gkiar:
Indeed, it would make more sense to fix the reader, I suppose :)  -- brainfart!

Thanks

tools/python/boutiques/localExec.py
```diff
@@ -1225,7 +1225,12 @@ def _findDOI(self, userIn):
             # Most recent DOI in file if user is publisher
             # Include check to ensure descriptor is unmodified
             if self.desc_dict.get('doi') is not None:
-                doi = self.desc_dict.get('doi')
+                # Zenodo returns the full DOI, but for the purposes of
+                # Boutiques we just use the Zenodo-specific portion (as its the
+                # unique part). If the API updates on Zenodo to no longer
+                # provide the full DOI, this still works because it just grabs
+                # the last thing after the split.
+                doi = self.desc_dict.get('doi').split('/')[-1]
```
glatard: This should be called `zenodo_id` but otherwise all good!

gkiar: Agreed :)

```suggestion
                zenodo_id = self.desc_dict.get('doi').split('/')[-1]
```

tools/python/boutiques/localExec.py
```diff
@@ -1225,7 +1225,12 @@ def _findDOI(self, userIn):
...
                 if loadJson(doi) == self.desc_dict:
```
```suggestion
                if loadJson(zenodo_id) == self.desc_dict:
```

tools/python/boutiques/localExec.py
```diff
@@ -1225,7 +1225,12 @@ def _findDOI(self, userIn):
...
                     return doi
```
```suggestion
                    return zenodo_id
```

tools/python/boutiques/localExec.py
gkiar: Actually, I should fix this in the puller, which actually is using the DOIs, since that's the place it fails. Everywhere else, we should use full DOIs or Zenodo IDs equivalently, and handle it appropriately in the `Puller`. Thoughts?

gkiar:
Actually, I should fix this in the puller, which actually is using the DOIs, since that's the place it fails. Everywhere else, we should use full DOIs or Zenodo IDs equivalently, and handle it appropriately in the Puller (even the `loadJson` utility uses the `Puller`). Thoughts?

gkiar:
@glatard ready for re-review?

Issue #397, opened 2019-02-07
Support basenames and wildcards in output files


gkiar:
Currently, in `output-files` we cannot use the `basename` of a parameter in substitution, or a `*` (wildcard).

Supporting wildcard seems easy - we just need to `glob` for the files after the templates are created.

For supporting `basename`s, we could add the following logic:
1. the `uses-basename` parameter added to `output-files`
2. when doing substitution, if this flag is set, all inputs of type `File` used in the `path-template` will be stripped to their basename before substitution (`path-template-stripped-extensions` will still be applied after this is done).
3. If you really want to half-use this feature, and only use the basename for one file and full path for another in the same output file name, you can make the input type of the file you'd like not to be reduced a `String` instead. This of course looses 1 step of input validation in that case at the cost of output validation, but is also a somewhat ridiculous use case that shouldn't come up often.

ramou:
For reference, this is the example I found for concisely applying wildcard searches that I think would be consistent with the above:
https://stackoverflow.com/questions/4296138/use-wildcard-with-os-path-isfile#4297946

glatard:
isn't it reasonable to assume that basenames will always be used when the value-key isn't a prefix of the template? And that when the value-key is a prefix, we use the value as is?

gkiar:
To be clear @glatard - you're saying if the `path-template` starts with some command-line key for a file, don't strip down to the basename, otherwise, always do? I support that.

glatard:
yes, that's what I mean. The rationale is that not stripping the basename should never hurt when the path is at the beginning of the template. And not stripping will always hurt when the path is in the middle of the template.

ramou:
Would it be weird to ask for examples of the different ways it could go and how it is expected that boutiques should interpret it?

glatard:
Sure, say that we have a value-key [FILE] mapped to value "/home/foo/file.txt", with stripped extensions containing "txt", then:
- `[FILE]_result.gz` would resolve to `/home/foo/file_result.gz`
- `result-[FILE].gz` would resolve to `result-file.gz`

gkiar:
Agreed. So this issue is just to implement that behaviour and fix wildcarding.

erinb90:
We might also want to exclude paths containing wildcards when validating that all output path templates are unique. For example, one of my descriptors produced from a Nipype interface had several output lists whose files all had extension `.nii`. So I put `*.nii` as the path template for all of them, but this doesn't validate for path template uniqueness.

[...]

PR #527, opened 2019-07-16
Improved output substitution

gkiar:
Addresses #397 

Previous behaviour:
- No truncation of values beyond stripping file extensions

Current behaviour:
- For `output-files` which contain `value-key` substitutions: inputs of `type="File"` used for these substitutions that do not *start* the `path-template` (i.e. `template.find(value-key)` does not return location `0`) will be modified to only insert the `os.path.basename` of the value, rather than the complete path.


glatard:
Looks great to me, thanks for doing this! I'm not sure if this PR closes #397 though, as wildcards weren't mentioned here.

Issue #521, opened 2019-06-28
Provenance publication on Nexus


glatard:
## Purpose
Extend `bosh data publish` to publish invocations to [Nexus](https://github.com/BlueBrain/nexus). Make it such that provenance captured by `bosh` can be linked to other provenance records, for instance produced by CBRAIN. 

#### Related issues
https://github.com/CONP-PCNO/conp-dataset/issues/52

## Context
`bosh data` provides a basic, minimalistic way to publish JSON objects representing execution provenance on Zenodo. In the context of [CONP](https://github.com/CONP-PCNO), we are aiming at consolidating provenance in a consistent repository. Nexus is being tested for this purpose (see https://github.com/CONP-PCNO/conp-dataset/issues/52). The goal here would be that records stored by `bosh data` can be published in Nexus, and extended in a way that other platforms can link to them. This stems from a discussion with @shots47s. 

## Preferred solution

1. Add an option to `bosh exec launch` to append an arbitrary JSON object to the provenance record. For instance, `bosh exec launch <tool_id> <invocation> --provenance '{ "engine": "http://cbrain.ca", "dataset id": "1234", "cluster": "beluga" }'` would add the object following `--provenance` to the provenance record. Here is an example of a hypothetical provenance record produced from this execution:
```
{
    "summary": {
        "name": "fslmaths",
        "descriptor-doi": "descriptor_fslmaths_2019-06-10T02:27:32.132502.json",
        "date-time": "2019-06-10T02:27:36.835733"
    },
    "public-invocation": {
        "input_file": {
            "file-name": "test.nii.gz",
            "md5sum": "433b12536427334ded8e10eeb4a62d00"
        },
        "bin": true
    },
    "public-output": {
        "stdout": "",
        "stderr": "",
        "exit-code": 0,
        "error-message": "",
        "shell-command": "fslmaths  ./test.nii.gz                                     -bin                                                                               ./output.nii.gz ",
        "missing-files": {},
        "output-files": {
            "output_file": {
                "file-name": "output.nii.gz",
                "md5sum": "7547190d2f8484f1b48e8d8b6cde44eb"
            }
        }
    },
   "additional-information": {
           "engine": "http://cbrain.ca",
           "dataset id": "1234",
           "cluster": "beluga"
   }
}
```

[...]

PR #532, opened 2019-07-30
Publish to nexus and add provenance record

joeyzhou98:
Extend publishing functions to allow publishing to Nexus.
See #521


joeyzhou98:
Hi @glatard, Nexus is organized in such a way so that there are organizations that can contain multiple projects, and in turn, those projects can contain multiple resources that have types like resolvers, schemas, views or files. This is described here: https://bluebrainnexus.io/docs/api/1.0/index.html. I have a few questions in term of the logistics of publishing to our Nexus server:
1) Should the deciding factors of publishing to which Nexus organizations and projects be manual, automatic or hard-coded? Manual would imply like specifying through the command line which organization and project to publish to, for example `bosh data publish --nexus --organization boutiques --project SIMON-dataset`. Automatic would imply reading the dataset/file title and creating/updating projects with that title so that you won't need to specify the project/organization, but I will need to know where to read the organization and project names from. Hard-coded would mean publishing to a specific organization and project every time.
2) Under what type of resources, mentioned above, should the files uploaded be a type of? Again we could make it so the code automatically decides based on some property of item in the files and I would need to know what are those properties, but again, specifying manually through flags is also an option

glatard:
Hi @joeyzhou98, sorry for the delay responding to this.
1. I think we should treat organization and projects as publication parameters, similar to `zenodo-access-token`. So it should be possible to store them in a config file or to pass them on the command line. The command line should have precedence over the config file (if no param is passed on the command line then read config file). You should reuse the same logic as the one used for `zenodo-access-token`.
2. Data sounds like the right type to use.

tools/python/boutiques/bosh.py
```diff
@@ -125,6 +126,9 @@ def execute(*params):
         parser.add_argument("--skip-data-collection", action="store_true",
                             help="Skips execution data collection and saving"
                             "to cache.")
+        parser.add_argument("--provenance", action="store", type=json.loads,
+                            help="Append provenance object to the "
```
glatard: Should be `Append JSON object to the generated record.`

tools/python/boutiques/nexusHelper.py
```diff
@@ -0,0 +1,108 @@
...
+        json_creds[self.config_token_property_name()] = access_token
+        with open(self.config_file, 'w') as f:
+            f.write(json.dumps(json_creds, indent=4, sort_keys=True))
+        if self.verbose:
+            print_info("Nexus access token saved in {0}".
+                       format(self.config_file))
+
+    def get_nexus_endpoint(self):
+        # change once nexus is setup
+        endpoint = "https://sandbox.bluebrainnexus.io/v1" if self.sandbox \
```
glatard: Could you create an issue for that?

joeyzhou98:
Hi, @glatard this is a big PR, it might take some time to review everything

tools/python/boutiques/nexusHelper.py
```diff
@@ -0,0 +1,155 @@
...
+        json_creds["nexus-project"] = project
+        with open(self.config_file, 'w') as f:
+            f.write(json.dumps(json_creds, indent=4, sort_keys=True))
+        if self.verbose:
+            print_info("Nexus access token, organization and project"
+                       " saved in {0}".format(self.config_file))
+
+    def get_nexus_endpoint(self):
+        # change once nexus is setup
+        endpoint = "https://sandbox.bluebrainnexus.io/v1" if self.sandbox \
```
glatard: Could you mention this URL in the helptext so that users know where to create an account?

PR #533, opened 2019-07-31
Zenodo search query length check

xlecours:
This fixes the query string in searcher when an empty string was passed to the function. 
Since `''` is different than `None` the empty string remained and the request query string was `**~ resulting in something like `...?q=**&keyword...`. A 500 Error was returned by Zenodo.


xlecours:
will send to develop branch

Issue #535, opened 2019-08-04
bosh search with "boutiques" as the query term returns a malformed dictionary


shots47s:
Running bosh search "boutiques" will produce an error due to a value that is return that doesn't have a key for ['stats']['version_views']

## Expected Behavior

running 
```
bosh search "boutiques" 
```
Should return all of the descriptors in Zenodo.

## Current Behavior
```
Traceback (most recent call last):
  File "/Users/stbrown/conp-portal/venv/bin/bosh", line 11, in <module>
    load_entry_point('boutiques==0.5.20.post1', 'console_scripts', 'bosh')()
  File "/Users/stbrown/conp-portal/venv/lib/python3.6/site-packages/boutiques/bosh.py", line 702, in bosh
    out = search(*params)
  File "/Users/stbrown/conp-portal/venv/lib/python3.6/site-packages/boutiques/bosh.py", line 466, in search
    return searcher.search()
  File "/Users/stbrown/conp-portal/venv/lib/python3.6/site-packages/boutiques/searcher.py", line 53, in search
    return self.create_results_list(results.json())
  File "/Users/stbrown/conp-portal/venv/lib/python3.6/site-packages/boutiques/searcher.py", line 70, in create_results_list
    (id, title, description, downloads) = self.parse_basic_info(hit)
  File "/Users/stbrown/conp-portal/venv/lib/python3.6/site-packages/boutiques/searcher.py", line 125, in parse_basic_info
    downloads = hit["stats"]["version_downloads"]
KeyError: 'version_downloads'
```
## Possible Solution (optional)
Need make the parsing of the dictionaries returned more robust to malformed entries.

## Steps to Reproduce

1. bosh search "boutiques"
1.
1.
1.

PR #536, opened 2019-08-04
Added a catch on version_downloads key

shots47s:
It appears if the tool…has never been downloaded, this doesn't exist rather than returning 0.

Fixes issue #535

### DIFF
--- a/tools/python/boutiques/searcher.py
+++ b/tools/python/boutiques/searcher.py
@@ -13,7 +13,7 @@ class Searcher():
 
     def __init__(self, query, verbose=False, sandbox=False, max_results=None,
                  no_trunc=False, exact_match=False):
-        if len(query or '') > 0:
+        if query is not None:
             self.query = query
         else:
             self.query = 'boutiques'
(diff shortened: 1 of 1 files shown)


tools/python/boutiques/searcher.py
```diff
@@ -13,7 +13,7 @@ class Searcher():
 
     def __init__(self, query, verbose=False, sandbox=False, max_results=None,
                  no_trunc=False, exact_match=False):
-        if len(query or '') > 0:
+        if query is not None:
```
glatard: Not sure why this is reverted: it was implemented in https://github.com/boutiques/boutiques/pull/534/files

tools/python/boutiques/searcher.py
```diff
@@ -121,7 +121,9 @@ def parse_basic_info(self, hit):
         id = "zenodo." + str(hit["id"])
         title = hit["metadata"]["title"]
         description = hit["metadata"]["description"]
-        downloads = hit["stats"]["version_downloads"]
+        downloads = 0
+        if "version_downloads" in hit["stats"]:
+            downloads = hit["stats"]["version_downloads"]
```
glatard: Not a big deal but it would be more Pythonic to write `downloads = hit[..][...] if "version_downloads" in hit["stats"] else 0`.

PR #537, opened 2019-08-07
suppressed annoying "docker is ..." message

gkiar:


tools/python/boutiques/localExec.py
```diff
@@ -491,7 +491,7 @@ def _cleanUpAfterSingPull(self, lockDir):
             del os.environ["SINGULARITY_PULLFOLDER"]
 
     def _isDockerInstalled(self):
-        return not subprocess.Popen("type docker", shell=True).wait()
+        return not subprocess.Popen("type docker 1>/dev/null", shell=True).wait()
```
```suggestion
        return not subprocess.Popen("type docker 1>/dev/null",
                                    shell=True).wait()
```

Issue #392, opened 2019-02-05
Export descriptor from docopt


glatard:
http://docopt.org

gkiar:
Yeah - pretty print is using the standard argparse formatting for parameters, so when we support one of creating descriptors from docopt or pretty print, we get the other essentially free

edickie:
ps we found out this is a little harder than we thought. Would be great though! pretty please!

glatard:
Hi @edickie could you post a summary of what you tried and why it turned out to be harder than expected?

edickie:
We thought we could convert from docopt to an argparse object but it turns out no..so now I'm back to writing the .json from scratch

glatard:
Hi @edickie, could you provide an example of a docopt script to be converted?

edickie:
Thanks. I have the draft boutiques bosh for the bidsapp and the script in
this branch

https://github.com/edickie/ciftify/tree/enc_addboutique_spec/ciftify/bidsapp


I would love to draft more boutiques spec around most of the stuff in the
bin/ for this project. (it's all docopt).
https://github.com/edickie/ciftify/tree/enc_addboutique_spec/ciftify/bin

DarrinFong:
Hi, I don't quite understand this issue after having read about docopt. My interpretation is that we want to use docopt's parser instead of argparse's to generate a descriptor json file. 

May I get an example of the command to run which requires some parsing by argparse? It seems like just running ```bosh create x.json``` generates a default sample descriptor.

glatard:
Your interpretation is correct: we want to be able to export a descriptor from a Python script that has its input/output parameters described with docopt. 

Similarly to argparse, export from docopt should work from a Python code, not from the command line. The example usage from README.md is:
```
import boutiques.creator as bc
newDescriptor = bc.CreateDescriptor(myparser, execname="/command/to/run/exec")
newDescriptor.save("my-new-descriptor.json")
```
Currently `myparser` has to be an `argparse` parser. `bc.CreateDescriptor` should work similarly with some docopt object.

glatard:
Just chatted with @gkiar about this issue. It would be nice to convert the docopt object to an argparse parser, so that `CreateDescriptor` could work consistently for both formats. There might be existing code to convert from docopt to argparse, btw.

edickie:
That would be great! My reading of the docopt code made me think it might
be harder to do. But I'm excited with there is already a docopt-to-argparse
that might work.

PR #539, opened 2019-09-04
Requires-inputs for groups

DarrinFong:
### Descriptions
Inputs' requires-inputs now allows group id, such that an input is only valid if a member of the aforementioned group is present (for mutually-exclusive and one-is-required groups).

### Before changes
The following use case could only be constrained in a way such that ```a -> X``` and ```b -> X```,
X can't require A OR B. Which makes ```toolFOO X``` a false positive valid statement.
```
toolFOO X a
toolFOO X b
```

### After changes
The following example of a simplified descriptor describes the same use case as mentioned above, and it ensures ```X -> A ⊕ B```:
```
inputs: [
    {"id": X [...], requires-inputs: ["A_B"]},
    {"id": A [...], requires-inputs: ["X"]},
    {"id": B [...], requires-inputs: ["X"]},
],
groups: [
    "id": A_B [...],
    "members": ["A", "B"]
]
```
#### This feature is required for the complete implementation of: #392


tools/python/boutiques/localExec.py
```diff
@@ -701,6 +722,27 @@ def isOrCanBeFilled(targ):
                     return False
             return True
 
+        # For optional params, don't add if it requires inputs not
+        # already chosen. Requires-complete can be satisfied but is
+        # too complicated to check without building dependency tree.
+        # (Added after requires-inputs: group was implemented)
+        def groupMemberCanBeAdded(targ):
```
glatard: Add test of this method and the other additions to simulate/example.

Issue #540, opened 2019-09-25
Mounting several mount points 


prioux:
Bosh fails to launch a tool in a singularity container when several mount points are asked to be mounted.

The bug was introduced in:

https://github.com/boutiques/boutiques/commit/0873f43f93637ae0de49a6811dee96d051f142fa

Both the original code and the modified code are wrong.

When several mountpoints must be added to the singularity launch command, distinct `-B mountpoint` options must be added to the command line (e.g. `-B abcd -B efgh -B ijkl`).

The original code built `abcd -B efgh -B ijkl` (notice the missing leading `-B`).

The new code builds `-B abcd efgh ijkl` (notice the missing `-B`s for the second and third mountpoints).

natacha-beck:
Closed by https://github.com/boutiques/boutiques/commit/4830d21a57a1ffbe558b8b0a7e0d6603888b41bf

prioux:
Great! I'm 4 months late!

PR #541, opened 2019-09-26
Dev

natacha-beck:
Should fix the issue #540 .

And add comments about the tricky part, the code expects the presence of at least one item in the mounts_strings.


tools/python/boutiques/localExec.py
```diff
@@ -279,40 +279,50 @@ def execute(self, mount_strings):
...
+
             if conTypeToUse == 'docker':
                 envString = " "
                 if envVars:
                     for (key, val) in list(envVars.items()):
                         envString += " -e {0}='{1}' ".format(key, val)
+
                 # export mounts to docker string
+                # Warning: mount_strings should have at least one entry ()
+                # otherwise docker run will failed
```
glatard: It should never happen but could we assert that this is true, just in case?

tools/python/boutiques/localExec.py
```diff
@@ -279,40 +279,50 @@ def execute(self, mount_strings):
...
-                                     singularity_mounts +
-                                     ' -W ' + launchDir + ' ' +
-                                     conOptsString +
-                                     str(conPath) + ' ' + dsname)
+                # Warning: mount_strings should have at least one entry ()
+                # otherwise singularity run will failed
+                singularity_mounts = ' -B '.join(mount_strings)
+                container_command  = (envString               + 'singularity exec ' +
+                                     '--cleanenv '            +
+                                     ' -B '                   + singularity_mounts  +
```
glatard: I'm not sure how that fixes anything, you removed "-B" from l318 and put it back here.

natacha-beck: It was tested... But just realize that your dev branch has the fix for the issue #540.

natacha-beck:
Already fix in dev. I close it.

Issue #542, opened 2019-09-26
Bug in `get_entry_point` when use `bids import`


natacha-beck:
`bosh import bids` failed to retrieve proper entry point according to the following Dockerfile: 
`https://github.com/BIDS-Apps/oppni/blob/master/Dockerfile`

The definition of the ENTRYPOINT is:

 `ENTRYPOINT ["python", "/oppni/oppni.py"]`

note the space between the 2 elems of the ENTRYPOINT definition. It will make the function return  `python",` that is not what it is suppose to be. 
 
## Expected Behavior
In mean time the definition of the ENTRYPOINT seems to be wrong in this Dockerfile according to Docker documentation that specify the following: 

> ENTRYPOINT ["executable", "param1", "param2"]

Then it should be the following:
> ENTRYPOINT ["python /oppni/oppni.py"]

Then my ENTRYPOINT in the Boutiques descriptor became `python` instead of `python  /oppni/oppni.py`  

## Current Behavior
See description at the top.

## Possible Solution (optional)
Refatoring `get_entry_point` to split properly

## Steps to Reproduce
1. git clone https://github.com/BIDS-Apps/oppni
1. bosh import bids oppni.json oppni/

natacha-beck:
Do Boutiques expect to have only `executable` define in ENTRYPOINT or it is possible to get params define in ENTRYPOINT ?

gkiar:
It's possible to get params, as well - in this case, `python` being the executable, `/path/to/oppni` as the second, and whatever args as `2+`. This appears to indeed be a bug in how we're parsing entrypoints, and we should ideally support both formats (i.e. separation by spaces and multiple list items).

You said you have a fix ready to roll in, @natacha-beck

PR #543, opened 2019-09-27
Fix for entrypoint handling

natacha-beck:
Should fix the following #542

### DIFF
--- a/tools/python/boutiques/importer.py
+++ b/tools/python/boutiques/importer.py
@@ -99,9 +99,25 @@ class Importer():
         with open(os.path.join(self.input_descriptor, "Dockerfile")) as f:
             content = f.readlines()
         for line in content:
-            split = line.split()
-            if len(split) >= 2 and split[0] == "ENTRYPOINT":
-                entrypoint = split[1].strip("[]\"")
+            if line.startswith("ENTRYPOINT"):
+                entrypoint_values = line.strip("ENTRYPOINT").strip()
+
+                # According to DOCKER documetation the ENTRYPOINT can
+                # be like:
+                #   > ENTRYPOINT ["executable", "param1", "param2"] (exec form, preferred)
+                # or:
+                #   > ENTRYPOINT command param1 param2 (shell form)
+                if not entrypoint_values:
+                    entrypoint = None
+                # ENTRYPOINT ["executable", "param1", "param2"] (exec form, preferred)
+                elif entrypoint_values.startswith("["):
+                    entrypoint_values = entrypoint_values.strip("[]")
+                    entrypoint = entrypoint_values.split(",")[0].strip()\
+                        .strip("\"")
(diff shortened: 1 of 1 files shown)


tools/python/boutiques/importer.py
```diff
@@ -99,9 +99,30 @@ def get_entry_point(self, input_descriptor):
...
-                entrypoint = split[1].strip("[]\"")
+            if line.startswith("ENTRYPOINT"):
+                entrypoint_values = line.strip("ENTRYPOINT").strip()
+
+                # According to DOCKER documetation the ENTRYPOINT can
+                # be like:
+                #   > ENTRYPOINT ["executable", "param1", "param2"]
+                # or:
+                #   > ENTRYPOINT command param1 param2
+                if not entrypoint_values:
```
gkiar: Thanks, @natacha-beck ! Just in the interest of perhaps making it a bit more readable and Pythonic, we might be able to simplify the logic to the following. The suggestion would replace everything up to line 126. What do you think?

```suggestion
                if not entrypoint_values:
                    return None

                if entrypoint_values.startswith("["):
                    # Replacing entrypoint_values with the Python-interpreted list version
                    entrypoint_values = json.loads(entrypoint_values)
                    # Adding single quotes around list items with spaces therein
                    entrypoint_values = [ev
                                         if " " not in ev else
                                         "'{0}'".format(ev)
                                         for ev in entrypoint_values]
                    return " ".join(entrypoint_values)

                return entrypoint_values
```

tools/python/boutiques/importer.py
```diff
@@ -99,9 +99,43 @@ def get_entry_point(self, input_descriptor):
...
+                    entrypoint_values = json.loads(entrypoint_values)
+                    # Adding single quotes around list items with spaces therein
+                    entrypoint_values = [ev
+                                         if " " not in ev else
+                                         "'{0}'".format(ev)
+                                         for ev in entrypoint_values]
+                    return " ".join(entrypoint_values)
+
+                return entrypoint_values
+                    entrypoint = None
```
gkiar: Hi @natacha-beck - this and the following lines in this block now need to be deleted. Thanks!

natacha-beck: @gkiar should be ok now.

glatard:
Hi @natacha-beck, thanks for working on this! The tests are currently failing due to pycodestyle (see Travis):
```
boutiques/importer.py:114:81: E501 line too long (90 > 80 characters)

The command "pycodestyle --max-line-length=80 boutiques/*.py boutiques/tests/*.py" exited with 1.
```

glatard:
Hi @natacha-beck, the tests are now failing following the release of CentOS8 last week. We fixed that issue with @DarrinFong yesterday, would you like to pull `develop` in your branch so we can check that everything now passes?

glatard:
Merging, as the test failures should be fixed in develop.

PR #549, opened 2019-10-16
Conditional outputs#303

DarrinFong:
## Purpose

Enables Conditional-path-template in descriptor's output-files. CPT will contain a list of objects whose key represents a boolean expression (python syntax), the first true expression's corresponding path will be selected (top down). If the output-file is required, a `default` key is must be included in the list of CPT. 

PT and CPT are mutex properties within an output-file

## Current behaviour
```
"output-files": [{
  ... ,
  "path-template": "[PARAM1].txt"
}]
```
## New behaviour
```
"output-files": [{
  ... ,
  "path-template": "[PARAM1].txt"
}, {
  ... , 
  "optional": false,
  "conditional-path-template":
  [
    {"opt1 and (opt2 > 10)": "[OPT1]-thing_large.txt"},
    {"opt1":                 "[OPT1]-thing.txt"},
    {"opt2 > 10":            "[PARAM1]-thing_large.txt"},
    {"default":              "[PARAM1]-thing.txt"}
  ]
}]
```


glatard:
Hi @DarrinFong, tests are currently failing, would you like to have a look?

tools/python/boutiques/validator.py
```diff
@@ -122,6 +122,19 @@ def inById(i):
             else:
                 errors += []
 
+    if 'output-files' in descriptor:
+        # Verify output file with non-optional conditional file template
+        # contains a default path
+        msg_template = ("OutputError: \"{0}\": Non-optional output-file with "
+                        "conditional-path-template must contain "
+                        "\"default\" path-template.")
+        for outF in [o for o in descriptor["output-files"] if
```
gkiar: Could we add logic here verifying that the conditions being evaluated are valid? I.e. the `conditional-path-template` fields are either "default" or contain a value that corresponds to one of the input IDs?

Also, another thing to check, is that the "types" are compatible among comparisons. I.e.:

```python
number < 6
```
is not equivalent to 
```python
number < "6"
```
and which one is desired is entirely based on the datatype of `number.`

We may want to at least throw a warning if the type of the value being compared is not the same type as the input used in the comparison? The easiest way to do this may be to generate a random value of that type and perform the comparison, and then catch + return the error we encounter (if we encounter one)?

tools/python/boutiques/localExec.py
```diff
@@ -1047,6 +1047,68 @@ def _generateOutputFileNames(self):
...
+        # if 'conditional-path-template' in outputItem
+        # (key=conditions, value=path)
+        # Initialize file name with path template or existing value
+        for outputId in [x['id'] for x in self.outputs
+                         if 'conditional-path-template' in x]:
+
+            for boolObj in self.safeGet(outputId, 'conditional-path-template'):
+                # Surround expression chars with space
+                # Then split by space to isolate ids
+                boolExp = list(boolObj)[0]
```
gkiar: When turning a dictionary into a list, is it always ordered `['key', 'value']`?

tools/python/boutiques/localExec.py
```diff
@@ -1047,6 +1047,68 @@ def _generateOutputFileNames(self):
...
+                splitExp = "".join(
+                    [c if c.isalnum() or c == "_" else
+                     " {0} ".format(c) for c in boolExp]).split(" ")
+                parsedExp = []
+                for word in [word.strip() for word in splitExp]:
```
```suggestion
                for word in [word.strip() for word in splitExp if len(word) > 0]:
```

I'd do the above so that empty characters are thrown out here, before any processing.

tools/python/boutiques/localExec.py
```diff
@@ -1047,6 +1047,68 @@ def _generateOutputFileNames(self):
...
+                    # therefore expression is false
+                    elif word in all_ids:
+                        parsedExp = ["False"]
+                        break
+                    # Word is an expression char, just append it
+                    else:
+                        parsedExp.append(word)
+                # If expression is true, set fileName
+                # Stop checking (if-elif...)
+                if " ".join("{0}".format(w) for w in parsedExp) == "default":
```
gkiar: I may be reading this wrong, but aren't what you have written on #L1078 (`parsedExp = ["False"]`) and what you're checking here, (` .... == "default"`) incompatible, so defaults would never be added? You know what's going on in here much better than I, @DarrinFong, so please just double check and let me know :)

gkiar:
Hey @DarrinFong  - I added some comments. Overall, this looks awesome! Thanks so much :)

gkiar:
Hey @DarrinFong  - I added some comments. Overall, this looks awesome! Thanks so much :)

tools/python/boutiques/localExec.py
DarrinFong: Had to test it :sweat_smile:, only the keys are taken when a dict is turned into a list. The reason for ```[0]``` is because there's always just one key:val per boolean object (type dict). 

I modified the code to be more readable: ```templateKey = list(templateObj.keys())[0]```

tools/python/boutiques/localExec.py
DarrinFong: change will be included in next commit

tools/python/boutiques/localExec.py
DarrinFong:
```python3
if word in in_out_dict:
    # in_out_dict are k:v pairs that will be in the output of simulate
elif word in all_ids:
    parsedExp = ["False"]
    break
```
The elif block should only 'falsify' the expressions that contain an ID which is both:
- in descriptor's inputs/output-files IDs
- NOT in selected values for simulate
Which means the only way for ```'default'``` to be evaluated ```False``` is when an ID of name ```'default'``` is present in the descriptor's inputs/output-files IDs. (This raises a concern of whether we have to add ```'default'``` to the list of invalid IDs).

gkiar: Great - thank you for the clarification :)

I do think that maybe we want to add "default" as a restricted word in the schema so that it can't be used as an ID... @glatard, thoughts?

gkiar:
@DarrinFong - want to investigate why this is failing, and we can re-review once that is fixed? Thanks!

tools/python/requirements.txt
```diff
@@ -2,4 +2,6 @@ jsonschema==2.6.0
 pytest==3.5.0
 requests==2.18.4
 simplejson==3.13.2
-docopt==0.6.2
\ No newline at end of file
+docopt==0.6.2
+nexus-sdk; 
+python_version >= '3.5';
```
glatard: This will break installation from pypi for Python<3.5. The solution has to work with the following install command:
```
pip install -r requirements.txt
```
For Python<3.5.
The nexus line should be written as:
`nexus-sdk; python_version >= '3.5'`

DarrinFong:
@gkiar Hey Greg, can you take a look at the conditions validation you mentioned in a conversation above? 
The current validation is to recursively (by parentheses) replace the values neighbouring a valid operator (==, !=, >=, <=, >, <) by their types (Flag, Number, String, Path) and check if they are of the same type.

tools/python/boutiques/validator.py
```diff
@@ -58,6 +62,82 @@ def inById(i):
             return descriptor["inputs"][inputGet("id").index(i)]
         return {}
 
+    def conditionalExpFormat(s):
```
gkiar: This function appears to exist in both `localExec.py` and `validator.py`, so perhaps we could move it to a `utils` file and import it?

tools/python/boutiques/validator.py
```diff
@@ -58,6 +62,82 @@ def inById(i):
...
+        return cleanedExpression
+
+    def isValidConditionalExp(exp):
+        # Return the type of a conditional expression's substring
+        def getSubstringType(s):
+            s = s.strip()
+            if s in schema_types:
+                return s
+            elif re.search(r'^[0-9]+\.?[0-9]*$', s):
+                return "Number"
```
gkiar: An additional type check we could do since the schema has a boolean for `integer = true/false` is check in the comparison whether or not the input that is of type number is an int or a float, and then also allowing us to be a bit more specific here and additionally separate floats from ints.

tools/python/boutiques/validator.py
```diff
@@ -22,6 +23,9 @@ def validate_descriptor(json_file, **kwargs):
     """
     path, fil = op.split(bfile)
     schema_file = op.join(path, "schema", "descriptor.schema.json")
+    allowed_keywords = ['and', 'or', 'false', 'true']
```
gkiar: Because capitalization may be funny/distinct across representations (in particular `True` versus `true`), we may want to either a) add more capitalizations to this list, or b) convert everything to an upper/lowercase value when we compare. Essentially just an extra precaution to avoid the hole of evaluating a JSON `true` and a boolean Python `True` to be distinct.

tools/python/boutiques/validator.py
DarrinFong: Function moved to utils.py!

tools/python/boutiques/validator.py
DarrinFong: I think we can allow comparison between floats and integers. If an ```NUM1``` input with ```integer = false``` evaluates to 9.5, the conditional path template with the key ```NUM1 <= 10``` should still be selected. What do you think?

gkiar: Yeah - you're right. How about in the case where `integer = true`? I think you're right, but we can just open an issue for this later if we think it is worth bringing back up :) Thanks!

tools/python/boutiques/validator.py
DarrinFong: I opted for option b), just in case someone needs to write ```ANd``` :smile:

tools/python/boutiques/validator.py
DarrinFong: Okay, I still implemented the type check for Integer if we need it (ea90f62e426ae449e4ccb82de8c5570c7cbd17b0). I reverted it on the last commit because I thought it was too strict.

PR #552, opened 2019-10-30
Fixed #522

### LINKED ISSUE #522: The `makeblastdb` descriptor may be getting pulled from live zenodo during tests
`makeblastdb` is the second most commonly downloaded boutiques descriptor from Zenodo, in spite of it not being a likely common download candidate compared to the other posted BLAST descriptors. I used it for some test stuff, so I am suspicious that it is getting pulled for real, which is maybe not the desired behaviour (If we need to pull live, we should make another example project)

## Expected Behavior
Running the test suite should not increase the descriptor downloads for `makeblastdb` on Zenodo. Alternately, the downloads of another example descriptor should increase instead

## Current Behavior

`bosh search` show `makeblastdb` in second place rank for downloads. I bet running the full test suite and then doing the same command would show it going up.

## Possible Solution (optional)
Either use the sandbox Zenodo for whatever test is doing this or create a new example project for the purpose of this testing.

ramou:
---
name: makeblastdb
about: Changed the mock.patch to mock urlretrieve, which was actually used 
title: 'Bug fix for issue #522 '
labels: enhancement

---

## Related issues
 - The wrong thing was being mocked, and it wasn't being patched properly anyway.
 - There seemed to be a change in the current example-tool version that was causing some test failure, so I changed it to use the newest id.

## Checklist

- Since this is a fix in the tests,  not the core codebase, I can confirm the fixed tests
- I am dubious about the changes to the example descriptor...

## Purpose
- This fixes a test so that makeblastdb will stop getting inflated numbers on zenodo

## Current behaviour
The makeblastdb entry on zenodo is downloaded every time while testing.

## New behaviour
A mock descriptor for makeblastdb is now used.
 
- [ ] Yes
- [ x] No

## Implementation Detail
I switched to patching boutiques.puller.urlretrieve. I created a temp file with the desired content and called the actual urlretrieve on the local file to imitate the result of calling on a url. It gets cached just the same.

### DIFF
--- a/tools/python/boutiques/tests/boutiques_mocks.py
+++ b/tools/python/boutiques/tests/boutiques_mocks.py
@@ -18,7 +18,7 @@ class MockZenodoRecord:
 
 # Mock object representing the current version of Example Boutiques Tool
 # on Zenodo
-example_boutiques_tool = MockZenodoRecord(2634310, "Example Boutiques Tool",
+example_boutiques_tool = MockZenodoRecord(2644621, "Example Boutiques Tool",
                                           "", "https://zenodo.org/api/files/"
                                           "009cc264-76f6-459c-a824-"
                                           "5db6241a979f/example1_docker.json")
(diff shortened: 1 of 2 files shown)


glatard:
lgtm!

PR #554, opened 2019-11-04
Fixed #553

### LINKED ISSUE #553: Zenodo may have changed how their api receives queries
It looks like the api query interface is diverging from the web interface (which may also not work quite the way it used to. An ambiguous not-quite-error is given and yields a well-formed zero-result answer that was a bit confusing. This seems to break many things in boutiques.

## Expected Behavior
When we run `bosh search` for something, we should get results

## Current Behavior
When we run `bosh search` it says zero results, no matter what. Examining the json you might see what appears to be an error message, but it is not flagged as an error.

## Context (optional)
So, my gut says they're migrating the api away from how the web-based interface is doing stuff, or something silly like that, but I have nothing to back that up
I believe They've switched how they pass stuff to the query, in fact I see evidence that they are using three approaches, such that this feels like it's mid-stream and undecided: a) using the : notationfor keywords, b) using a keywords parameter and c) using the base elastic search query notation.
So if you click on a keyword in an entry it opens the search using the colon notation that works, e.g. https://zenodo.org/api/records/?q=keywords:"Boutiques" If you click on the keyword checkboxes in the search results e.g. https://zenodo.org/search?page=1&size=20&q=Boutiques%20AND%20blast&keywords=Boutiques and if you type Boutiques AND blast into the search box it gives you https://zenodo.org/search?page=1&size=20&q=Boutiques%20AND%20blast The last option has the desired result.

## Possible fix (optional)
We need to change things that use the api, specifically so that it makes better use of the api:
https://github.com/boutiques/boutiques/blob/ac1214010f6c4f88eb91ea8d2fb165a8ad619bab/tools/python/boutiques/searcher.py#L57-L60

It needs to be something more like:
```
        r = requests.get(self.zenodo_endpoint + '/api/records/?q=%s '
                         'AND keywords:Boutiques&file_type=json&type=software'
                         '&page=1&size=%s' % (self.query, 9999))
```

Not sure what to do about the schema/version stuff :\

ramou:
---
name: #553 
about: I changed the way the records api was queries in searcher
title: 'Bug fix for issue #553'

---

## Checklist

- [x ] **DO** Unit tests pass. /well, as much as they pass on my [dev docker](https://hub.docker.com/r/ramou/boutiques_test_env)

## Purpose
This replaces the old query approach with one that appears consistent with the very limited documentation (and it now works)

## Current behaviour
It seems that Zenodo updated things Oct. 30 (https://github.com/zenodo/zenodo/commit/9e95b2893c509bddc52a9ee8acc1dd24e72a721b) and I think this represents their having deployed a bunch of updates, some of which related to bringing their ElasticSearch up to date, which may have caused what we were doing previously to break.

## New behaviour
Things should now work as they used to.
 
- [ ] Yes
- [ x] No

## Implementation Detail
I swtiched to putting all the keyword queries inline according to the ElasticSearch query syntax. I moved the query-string formatting earlier in the code. I kinda don't like that and think maybe I should turn self.query into a list of terms, with the exact term just being the unsplit query and build query_line just before using it... but I have a meeting and this fixes an immediate problem well enough.

The biggest difference is that this does not send query terms to anything other than keyword fields. I don't know what was intended by the last version, but this is now more exact. It means it won't search authors, for example. If I had a better idea what was desired, we could do that, though I had trouble specifically accessing stuff other than the keywords field, but I think that's just a matter of me getting used to ES.

### DIFF
--- a/tools/python/boutiques/searcher.py
+++ b/tools/python/boutiques/searcher.py
@@ -17,14 +17,20 @@ class Searcher():
         if query is not None:
             self.query = query
             if not exact_match:
-                self.query_line = \
-                    ' AND keywords:(/.*%s.*/)' % self.query
+                terms = self.query.split(" ")
+                self.query_line = ''
+                for t in terms:
+                    self.query_line = self.query_line + \
+                        ' AND keywords:(/.*%s.*/)' % t
             else:
-                self.query_line = \
-                    ' AND keywords:(/%s/)' % self.query
+                self.query_line =\
+                        ' AND keywords:(/%s/)' % self.query
         else:
             self.query_line = ''
             self.query = ''
+
+        if(verbose):
+            print_info("Using Query Line: " + self.query_line)
 
         self.verbose = verbose
         self.sandbox = sandbox
--- a/tools/python/boutiques/searcher.py
+++ b/tools/python/boutiques/searcher.py
@@ -54,10 +54,12 @@ class Searcher():
 
     def zenodo_search(self):
         # Get all results
-        r = requests.get(self.zenodo_endpoint + '/api/records/?q=%s&'
-                         'keywords=boutiques&keywords=schema&'
-                         'keywords=version&file_type=json&type=software'
-                         '&page=1&size=%s' % (self.query, 9999))
+        r = requests.get(self.zenodo_endpoint + '/api/records/?q='
+                         'keywords:(/Boutiques/) AND '
+                         'keywords:(/schema-version.*/) AND '
+                         'keywords:(/.*%s.*/)&'
+                         'file_type=json&type=software&'
+                         'page=1&size=%s' % (self.query, 9999))
         if(r.status_code != 200):
             raise_error(ZenodoError, "Error searching Zenodo", r)
         if(self.verbose):
--- a/tools/python/boutiques/searcher.py
+++ b/tools/python/boutiques/searcher.py
@@ -14,7 +14,7 @@ class Searcher():
     def __init__(self, query, verbose=False, sandbox=False, max_results=None,
                  no_trunc=False, exact_match=False):
 
-        if self.query is not None:
+        if query is not None:
             self.query = query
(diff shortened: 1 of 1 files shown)


glatard:
Hi @ramaou, thanks for looking into this! This doesn't seem to be fixing the issue on my end though:
```
$ bosh search --verbose
[ INFO ] Using Query Line:  AND keywords:(/.*boutiques.*/)
[ INFO ] Using Zenodo endpoint https://zenodo.org
[ INFO (200) ] Search successful for query "boutiques"
[ INFO ] Showing 0 of 0 results.
```
Did I miss anything?

ramou:
Two things I note. First, from the tests in Travis (much clearer since my test env shows many more and I haven't had time to vet what's up with that) the intent is for searched terms to be everywhere, not just in keywords. I'll switch it from `keywords:(/.*boutiques.*/)` to `(/.*boutiques.*/)` which should do the trick and pass those three tests.

Secondly, and my phone may send a message to this effect eventually, I currently haven't got it acting in a case insensitive fashion yet. The usual regex tricks did not work, but I will tinker some more and see if there's something else I can send ES to make it do that.

ramou:
So far I've come up with what I think is an awful solution. Others have come up with this to get around this "feature" or ElasticSearch:
https://discuss.elastic.co/t/regexp-and-case-insensitive/1288

That's right, filter the terms and apply per-character case groups.

ramou:
It appears to be case sensitive, at least with the capitalized keyword. I
have not made the search regex case insensitive successfully yet

ramou:
Oh, I understand the test now... that's a pain. I'll have to rewrite it because it's using the simply grabbed query. Since we can't do that anymore I'll have to think of something...

ramou:
Props to @la10736 for a clever way to make mock work as I wanted:
https://stackoverflow.com/questions/28286630/add-extra-arguments-to-side-effects-in-python-mock

Issue #445, opened 2019-03-14
Use appveyor


gkiar:
For windows build environments

PR #561, opened 2019-12-03
Boutiques on Windows

arthursw:
---
name: Fix windows issues (paths and file names)
about: Remove colon in file names (invalid characters in windows), and make path homogeneous among platforms.
title: 'Windows bug fix'
labels: windows support
assignees: ''

---

## Checklist

- [x] **DO** Unit tests pass.

For now unit tests do not pass because of the error: `DescriptorValidationError: [ ERROR ] Additional properties are not allowed ('deprecated' was unexpected)` (some descriptors seem out of date?)

## Purpose

This PR makes paths homogeneous among platforms (tested on windows and linux only but should work on mac as well).

When used in file names,`datetime.datetime.now().isoformat()` is replaced by `datetime.datetime.now().strftime("%Y-%m-%d_%Hh%Mm%Ss%fms")` to avoid the colon (`:`) character which causes problem on Windows.

The command `type` seems to have a different purpose on Windows so to check if Singularity or Docker is installed the command `docker --version` or `singularity --version` is called instead.
I had to remove "&>/dev/null" from the command line 542 of localExec.py.

## Current behaviour

Works on Linux and Windows.

## New behaviour

Now also works on windows. 
 
- [ ] Yes
- [x] No


glatard:
Hi @arthursw, thank you so much for working on this! I noticed that you added several JSON or txt files that aren't part of the code, probably because they were generated by the tests: do you mind removing them? It might be the reason why the tests are failing.

arthursw:
@glatard ok, I removed the test files ; sorry about that!

glatard:
The tests are now passing and the code makes a lot of sense to me. I'm just surprised that we don't have more Linux-specific things (like '\n' or manual path concatenations) elsewhere in the code! I have started looking at Appveyor config to test the code on Windows (see #445) but didn't manage to have it working yet. As these modifications don't seem to break anything on Linux, I'd suggest we merge them and then work on Windows testing through Appveyor. @arthurmasson, if you have any experience with Appveyor, your help would be greatly appreciated!

ramou:
I'll take tomorrow to bludgeon this. If it survives, or I don't, thumbs up.

arthursw:
Actually, I didn't check that every features work on windows... I just focused on what is required for the medInria plugin. I don't have any experience with Appveyor but I can have a try if you need it.

PR #566, opened 2019-12-19
Disc validate dict#518

### LINKED ISSUE #504: bosh example generates flags with value False
`bosh example` generates flags with value `false` when only `true` is valid. For instance:
```
$ bosh example zenodo.3240521
{
    "input_file": "f_input_file_74.m",
    "w": false
}
```
This fails input validation:
```
$ bosh invocation zenodo.3240521 -i '{"input_file": "f_input_file_74.m", "w":false}'
[ ERROR ] False is not one of [True]

Failed validating 'enum' in schema[10]['properties']['w']:
    {'enum': [True]}

On instance['w']:
    False
```
This is because inputs of type "Flag" can only take value "true": when flags are not used, they are simply omitted.

## Expected Behavior
Values generated for inputs of type "Flag" should always be "true". When "false", the input should just be omitted.

## Current Behavior
"false" is occasionally generated for inputs of type "Flag", resulting in invalid invocations.

## Steps to Reproduce
Run ` bosh example zenodo.3240521` until a flag input with value "false" appears.

@DarrinFong

### LINKED ISSUE #518: Refactoring localExec._validateDict
A lot of code in  `localExec.py._validateDict` should be replaced by a single call to `bosh.invocation`. 

https://github.com/boutiques/boutiques/blob/8e9e0acc8b61f83f29ecf918cc7841d466ec181d/tools/python/boutiques/localExec.py#L1024

DarrinFong:
---
name: bug fix #518 (Refactor localexec._validateDict)
about: _validateDict function in localexec is redundant. Should use boutiques.invocation
title: 'Bug fix for issue #518'
---

## Purpose

localexec contains a sizeable portion of spaghetti code, notably _validateDict. It's redundant and poorly documented, so we should remove it.

## Current behaviour

__validateDict_ is called in two places in _localExec.py_: 
- To verify random parameters generated for bosh simulate/example
- To verify input file (invocation) of the executor object

## New behaviour

__validateDict_ no longer exists, and is replaced by boutiques.invocation
There's a validation added to __validateDict_ during the bug fix #504 (testParam (False) flag is set to true or otherwise omitted). This validation has now been removed along with __validateDict_.


DarrinFong:
One of the job failed but seems to be a travis error

Issue #511, opened 2019-06-09
bare metal execution


glatard:
## Purpose
Add a `--no-container` option to run the command line outside of any container.

## Context
It is not possible to run containers in a Binder notebook easily, because the notebook runs in a Docker container started without `--privileged`. Even Singularity doesn't work in these conditions. This flag would allow people to use `boutiques` in such contexts, provided that they install all the required tools on their own, which is easy to do in case of Binder. In particular, it would simplify the writing of [Neurolibre](http://neurolibre.conp.ca/) tutorials that relies on Boutiques tools.

## Preferred solution (optional)
When `--no-container` is used, everything should behave as usual, but command line execution should be done as if no container image had been specified.

## Alternatives solutions (optional)
The current [Boutiques tutorial on Binder](https://mybinder.org/v2/gh/boutiques/tutorial/master?filepath=%2Fnotebooks%2Fboutiques-tutorial.ipynb) pre-loads `~/.cache/boutiques` with modified descriptors where the container image was removed. This is a hack and shouldn't be recommended in the long run.

@camarasu

PR #582, opened 2020-01-17
Baremetal exec

DarrinFong:
---
name: Bare metal execution
title: Feature implementation #511 
---

## Purpose

Run command line of a tool outside of a container with ```--no-container```

## New behaviour

Example: ```bosh exec launch [descriptor] [invocation] --no-container``` 
```--no-container``` flag defaults to false.

## Implementation Detail

In _localExec.LocalExecutor.execute_, setting ```conImage``` to None and ```conIsPresent``` to False will run command locally.

### DIFF
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -144,6 +144,9 @@ def parser_executeLaunch():
                         "to cache.")
     parser.add_argument("--provenance", action="store", type=json.loads,
                         help="Append JSON object to the generated record.")
+    parser.add_argument("--no-container", action="store_true",
+                        help="Execution takes place outside of container,"
+                        + " tool is executed in command line.")
     force_group = parser.add_mutually_exclusive_group()
     force_group.add_argument("--force-docker", action="store_true",
                              help="Tries to run Singularity images with "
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: tools/python/boutiques/schema/examples/baremetal/test_baremetal.json [hunk 1]
```diff
@@ -0,0 +1,30 @@
+{
+    "command-line": "echo \"Bare metal execution\" > [FILENAME] && cat [FILENAME]", 
+    "container-image": {
+        "entrypoint": false, 
+        "image": "neurodata/ndmg:v0.0.41-1", 
+        "index": "index.docker.io", 
+        "type": "docker"
+    },
```
glatard: Could you confirm that this still works if `container-image` isn't present in the descriptor?

tools/python/boutiques/bosh.py
```diff
@@ -144,6 +144,9 @@ def parser_executeLaunch():
                         "to cache.")
     parser.add_argument("--provenance", action="store", type=json.loads,
                         help="Append JSON object to the generated record.")
+    parser.add_argument("--no-container", action="store_true",
+                        help="Execution takes place outside of container,"
+                        + " tool is executed in command line.")
```
glatard: Help text should be "Launch invocation on the host computer, with no container. If 'container-image' appears in the descriptor, it is ignored."

tools/python/boutiques/schema/examples/baremetal/test_baremetal.json
DarrinFong: I tried the following descriptor/flag combination with `bosh exec launch` and they all behaved as expected:
no container-image -> exec in command line
no container-image + --no-container -> exec in command line
w/ container-image -> exec in container
w/ container-image + --no-container -> exec in command line

Issue #520, opened 2019-06-26
Boutiques publish should not reorder the properties (or have option to keep the original order))


MontrealSergiy:
## Purpose
Keep the preferred ordering of arguments to adhere to the principle of the least user astonishment. While the conventional json wisdom is to order a long list of properties alphabetically, some devs, such as Pierre R. rely on handpicked order, say order of importance.

## Context
Zenodo publishing, and perhaps other commands reorder the properties of alphabetically. While ordering might be useful it is not what most users would expect, and obscure the fact that there is actual semantic change (added property).

## Preferred solution (optional)

I suggest to make reordering optional or remove completely 

## Possible Implementation (optional)

there is some param in json.load to keep ordering (in an ordered dictionary)

prioux:
Ah people talk about me behind my back!

Anyway, here's a nice example of a descriptor I hand-crafted with love:

https://github.com/aces/cbrain/commit/2a72a34c29?diff=unified

Knowing me, that thing would also be full of comments if JSON allowed any.

Issue #480, opened 2019-05-09
Improve layout of JSON created by 'bosh create'


prioux:
I just tried bosh create for the first time, I found the layout of the JSON not that great when I opened it up with an editor. The important fields are all over the place. I would have
like to have at the very top:

    "name": "perlvers", 
    "tool-version": "v9.0.0"
    "command-line": "use English;print \"I am Perl $PERL_VERSION\\n\";", 
    "shell": "/usr/bin/perl",
    "description": "Perl Script Within Boutiques", 

in approximately that order (ignore my values, I'm testing stuff).

Also, there is a spurious blank space after each and every comma.

glatard:
Indeed, the fields are dumpted in alphabetical order, which isn't great. We should fix that in create, maybe in other places too.

prioux:
Related to #520

PR #586, opened 2020-02-02
Improve desc invoc outputs

DarrinFong:
## Related issues

#480 #520 

## Purpose

Keep the preferred ordering of arguments to adhere to the principle of the least user astonishment. While ordering might be useful it is not what most users would expect, and obscure the fact that there is actual semantic change (added property).

## Current behaviour

Most descriptor/invocation outputs are alphabetically sorted.

## New behaviour

Keep the preferred ordering of arguments according to a template or the original descriptor. 
Keep ordering of invocations according to the inputs of a descriptor.

## Implementation Detail

Custom sort for descriptor: defaults to ordering given by _tools/python/boutiques/templates/ordered_keys_desc.json_
Custom sort for invocation: no default, all set to order by input descriptor.

### DIFF
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -18,7 +18,7 @@ from boutiques.localExec import ExecutorError
 from boutiques.exporter import ExportError
 from boutiques.importer import ImportError
 from boutiques.localExec import addDefaultValues
-from boutiques.util.utils import loadJson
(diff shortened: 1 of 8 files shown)


DarrinFong:
Tests passed on my Travis build, problem might have been on Zenodo's end. 
https://travis-ci.org/DarrinFong/boutiques/jobs/646577945

Issue #588, opened 2020-02-19
Support image names containing an index


glatard:
Various commands crash when the container image contains an index name, for instance:
```
$ bosh exec prepare zenodo.3267250
[ ERROR ] Could not pull Singularity image: 
 * Pull command: singularity pull --name "shub---aces-cbrain-containers-recipes-fsl_v6.0.1.simg.tmp" shub://shub://aces/cbrain-containers-recipes:fsl_v6.0.1
 * Error: FATAL:   While pulling shub image: failed to get manifest for: shub://shub://aces/cbrain-containers-recipes:fsl_v6.0.1: no response received from singularity hub
```

## Expected Behavior

Commands should not crash. 

## Current Behavior

Commands crash because index is appended to the command name even though it's already present in the image name.

## Possible fix

Support image names containing an index name: in this case, don't append the index name when it's detected to be already present in the image name. The validator should check that if an index is present in the image name, the index field is either not present, or is a prefix of the image name.

PR #600, opened 2020-03-18
Support image name with index

DarrinFong:
# Related issues

#588 

## Purpose

Various commands crash when the container image contains an index name

## Current behaviour

Commands crash because index is appended to the command name even though it's already present in the image name.

## New behaviour

Index in image name is matched with regex and `container-image`'s `index` value will not be prepended to image name.
Validator raises error if an index is found in `container-image`'s `image` and it doesn't match the `index` value.


DarrinFong:
Thanks! Good call to add a test for container image with index, there were a few bugs uncovered by the test :smiley:

Issue #578, opened 2020-01-15
Simulation handles optional values differently depending how it is run (noticeable in config files)


gkiar:
## Expected Behavior

All methods of simulation should produce essentially the same performance. In the current implementation of `bosh exec simulate` without the `-i` option, it appears as though a "corner" is being cut somewhere and the inputs are generated+populated into the simulation simultaneously, rather than generating the inputs in a single step and simulating the execution with them in a second. This packaging of the generation + insertion results in inconsistent handling of optional values.

## Current Behavior

There are two "one-line" ways to easily generate invocations and simulate a command-line for tools:

```bash
$ # This generates an invocation, saves it to tmpinvo.json, and then simulates the execution with it
$ bosh example desc.json > tmpinvo.json && bosh exec simulate desc.json -i tmpinvo.json
$
$ # This simulates the execution by generating the invocation on the fly and then using it directly
$ bosh exec simulate desc.json
```
Unfortunately, as observed when working with config files, their behaviour is not the same. In the case of the former (longer) command: we get expected behaviour. Optional values that have defaults are inserted into the configuration file using those values. In the case of the latter (shorter) command: values that are optional, even with defaults, are omitted from the configuration file and blank lines are placed in their stead. A minimum descriptor to illustrate this, and the two outputs, can be found below.

desc.json:
```json
{
    "name": "tool",
    "tool-version": "1",
    "description": "a tool",
    "command-line": "tool_cli [CONFIG_FILE]",
    "schema-version": "0.5",
    "inputs": [
    {
        "id": "value",
        "name": "value",
        "type": "String",
        "value-key": "[VALUE]",
        "value-choices": ["yes", "no"],
        "default-value": "no",
        "optional": true
    }
    ],
    "output-files": [
    {
        "id": "config_file",
        "name": "Configuration file",
        "value-key": "[CONFIG_FILE]",
        "path-template": "config.toml",
        "file-template": [
            "hi__",
            "value = [VALUE]",
            "__bye"
            ]
        }
    ]
}
```

Output from command 1:
```
hi__
value = no
__bye
```

Output from command 2:
```
hi__

__bye
```

gkiar:
Note: here is yet another one-liner for generating invocations and simulating them to get the expected output, and this one doesn't require writing a temporary file to disc:

```bash
$ bosh exec simulate desc.json -i $(bosh example desc.json | tr -d [:space:])
```

gkiar:
[...]

PR #605, opened 2020-04-16
Bosh simulate consistency patch

DarrinFong:
## Related issues

#578 

## Purpose

`bosh exec simulate` should yield consistent results regardless of an invocation's presence through option `-i`.

## Current behaviour

Descriptor's `optional: true` inputs with `default-value` are not added to simulated command-line output, but are added when an invocation (created by bosh example) is supplied.

## New behaviour

`optional: true` inputs with `default-value` are added to simulated command-line output consistently.

## Implementation Detail

In `simulate`, `localExec.addDefaultValues` is called before `LocalExecutor.generateRandomParams`.

`LocalExecutor.generateRandomParams` no longer takes argument `n` for the number of iterations to attempt. This is to prevent the `LocalExecutor.in_dict` from clearing before each iteration (`in_dict` contains inputs to use when generating random param values).

### DIFF
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -282,7 +282,7 @@ def execute(*params):
             # Add optional inputs with default-value to inputs_dict,
             # which is then populated with random params
             executor.in_dict = addDefaultValues(executor.desc_dict, {})
-            executor.generateRandomParams()
+            executor.generateRandomParams(generateCmdLineFromInDict=True)
 
         if results.json:
             sout = [json.dumps(
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -279,7 +279,10 @@ def execute(*params):
                                   "requireComplete": results.complete,
                                   "sandbox": results.sandbox})
         if not inp:
-            executor.generateRandomParams(1)
+            # Add optional inputs with default-value to inputs_dict,
+            # which is then populated with random params
+            executor.in_dict = addDefaultValues(executor.desc_dict, {})
+            executor.generateRandomParams()
 
         if results.json:
             sout = [json.dumps(
@@ -674,7 +677,7 @@ def example(*params):
                               "skipDataCollect": True,
                               "requireComplete": results.complete,
                               "sandbox": results.sandbox})
-    executor.generateRandomParams(1)
+    executor.generateRandomParams()
     return json.dumps(
         customSortInvocationByInput(executor.in_dict, descriptor), indent=4)
 
--- a/tools/python/boutiques/localExec.py
+++ b/tools/python/boutiques/localExec.py
@@ -938,7 +938,7 @@ class LocalExecutor(object):
         # encountered before blowing up
         except Exception as e:  # Avoid BaseExceptions like SystemExit
             sys.stderr.write("An error occurred in validation\n"
-                                "Previously saved issues\n")
+                             "Previously saved issues\n")
             for err in self.errs:
                 sys.stderr.write("\t" + str(err) + "\n")
             raise e  # Pass on (throw) the caught exception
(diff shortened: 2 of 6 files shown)


glatard:
@gkiar would you like to review and test this one?

gkiar:
@glatard @DarrinFong yep - sorry for the delay, I'll get to this tomorrow :)

gkiar:
@DarrinFong the changes look good, thanks!

However, I'm still getting the exact same error in the config file case as is documented in #578 . Could you add the descriptor I pasted in that issue to the test suite, and verify consistent performance? Thanks for the hard work!

DarrinFong:
> However, I'm still getting the exact same error in the config file case as is documented in #578
@gkiar Hi Greg, I couldn't replicate the same errors despite using the descriptor and commands in #578 . Can you let me know if the new test covers the issue? Thanks a lot!

gkiar:
Hmm @DarrinFong, could it perhaps be the way the command-line utility is invoking this? I've just checked out your last commit and did/got the following:

```bash
[DarrinFong/bsimulate_inconsistency_#578] gkiar $ git log -1
commit d986441e55d665453831e25424f2d43401620e6b (HEAD -> DarrinFong/bsimulate_inconsistency_#578)
Author: darrinfong <darrinfong@gmail.com>
Date:   Tue Apr 28 16:45:59 2020 -0400

    python string incompatibility fix
[DarrinFong/bsimulate_inconsistency_#578] gkiar $ cat tmp.json
{
    "name": "tool",
    "tool-version": "1",
    "description": "a tool",
    "command-line": "tool_cli [CONFIG_FILE]",
    "schema-version": "0.5",
    "inputs": [
    {
        "id": "value",
        "name": "value",
        "type": "String",
        "value-key": "[VALUE]",
        "value-choices": ["yes", "no"],
        "default-value": "no",
        "optional": true
    }
    ],
    "output-files": [
    {
        "id": "config_file",
        "name": "Configuration file",
        "value-key": "[CONFIG_FILE]",
        "path-template": "config.toml",
        "file-template": [
            "hi__",
            "value = [VALUE]",
            "__bye"
            ]
        }
    ]
}

[DarrinFong/bsimulate_inconsistency_#578] gkiar $ bosh exec simulate tmp.json && cat config.toml
Generated Command:
tool_cli config.toml
hi__
value = no
__bye

[DarrinFong/bsimulate_inconsistency_#578] gkiar $ bosh example tmp.json && cat config.toml
{}
hi__

__bye
```

DarrinFong:
@gkiar If I remember correctly `bosh example` is behaving as expected (#373), the `value` field will be populated in the config file if `--complete` flag is used. However the default value will not be used (value is randomly generated amongst `value-choices` in localExec). I'm not sure where to go from here, maybe have a new issue where `bosh example [desc] --complete` would prioritise `default-value`?

gkiar:
Hey @DarrinFong I think you're right in the case of common descriptors/invocations, but I think in a config file the default values should always be populated as having empty lines in a config file where some input is expected may cause issue. In this case, I think it's safer to assume the value isn't a default in the sense of "will be assumed" but rather as "the expected value" of a parameter.

What do you think? Could you maybe build in a switch that automatically forces `--complete` for all arguments which would get substituted within a config file, and leave the behaviour for other arguments as-is?

In any case, I think we should have consistency between our simulation methods -- I wouldn't expect `bosh exec simulate` and `bosh example` do give us different behaviour, and think harmonizing these is important.

Thanks for the hard work on this!

DarrinFong:
Hey @gkiar, I agree we should expect consistent behaviour, so I went back and thought about this: The purpose of `example` and `exec simulate` should be different since `example` is to generate a sample invocation, and `exec simulate` is to simulate an execution.

From that, we shouldn't expect `example` to generate a config file at all, because the file should only be generated during an execution. Therefore I propose to remove any output other than the invocation from `example` and leave `exec simulate` unchanged from last commit.

Let me know if I made an incorrect assumption on `example`'s or `exec simulate`'s behaviour. Thanks for the followup :smiley:

gkiar:
@DarrinFong -- I agree completely, great point and yes let's make that happen! Thanks for the great work on this, it's hugely appreciated!

gkiar:
Looks great :) Nicely done, @DarrinFong !

Issue #463, opened 2019-04-10
bosh test does not use imagepath, though it uses exec launch


ramou:

ramou:
When writing something to systematically test descriptors it is sensible to purge each test environment, but to take advantage of previously pulled images. I had expected to find this in place, but it is likely just one of those things that hadn't previously come up.

PR #608, opened 2020-04-20
Btest use imagepath

DarrinFong:
## Related issues

#463 #5 

## Purpose

Enable imagepath for `bosh test` to speed up testing within a tool.

## Current behaviour

`bosh test` uses `bosh exec launch` but does not have the `--imagepath` option enabled.

## New behaviour

`--imagepath` option is added to `bosh test` so the image path is passed to `bosh exec launch` during individual tests.

## Implementation Detail

During `bosh test`, additional options such as `--skip-data-collection` and `--imagepath` are added to a dict and passed to `test` in test.py. These additional options are appended to launch arguments.

### DIFF
--- a/tools/python/boutiques/bosh.py
+++ b/tools/python/boutiques/bosh.py
@@ -570,6 +570,9 @@ def parser_test():
     parser.add_argument("--sandbox", action="store_true",
                         help="Get descriptor from Zenodo's sandbox instead of "
                         "production server.")
+    parser.add_argument("--imagepath", action="store",
+                        help="Path to Singularity image. "
+                        "If not specified, will use current directory.")
     return parser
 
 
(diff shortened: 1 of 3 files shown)


DarrinFong:
@ramou Hi Stuart, do you have a descriptor with tests that use a singularity image? I want to make sure the imagepath is used during tests. Thanks!

ramou:
Hey Darrin, not sure if I understand what you're looking for, but I have:
https://github.com/bioinformatics-group/BoutiquesDescriptors/blob/master/ncbi-blast/2.7.1/blastp.json
with
https://github.com/bioinformatics-group/BoutiquesDescriptors/tree/master/ncbi-blast/2.7.1/blastp

glatard:
@ramou do you think you would be able to review this and check that it fixes what your initial issue was about?

ramou:
Hi, sorry that I didn't make the initial connection when I asked my question.

I've run through and confirmed that it's now passing the `--imagepath` on to `pytest` to use the same way it works with exec... playing with it, I see I have to change my approach, but not in a bad way (I'd hoped if you passed it a directory it'd name the file whatever.simg based on what it would normally be, but of course that's problematic, so its best to let the caller decide what the path should be if there'll be conflicts). 

When chaining tests in BoutiquesDescriptor I'll just pass an image name, and if they use the same image, it won't load a new image name... I should probably pass a param to hold the images there instead of re-loading them every call for dev purposes... but this is a ticket for https://github.com/bioinformatics-group/BoutiquesDescriptors once I clean it up and migrate it under the boutiques project.

Thanks, Darren!

DarrinFong:
Hi @ramou, I just want to do check if this issue's ready to be merged or there are some changes I need to make. Thanks!

ramou:
Hi @DarrinFong, it's fine like this.

Issue #575, opened 2020-01-15
"estimate" descriptor input fields from config file template


gkiar:
## Purpose

Given a configuration file for a tool with default values, generate a base for the descriptor including the inputs and config file template. This would of course just be an estimate because it would be difficult/impossible to identify which strings are just that versus a selection from an enum.

## Context

We have a similar tool for generation of descriptors from argparse and docopt; config files (sticking to the very standardized `.toml`, `.yml`, and `.json` formats) could be similarly handled.

glatard:
We could also generate inputs from a command-line template. This should be added to `bosh create`.

DarrinFong:
Hi @gkiar, my understanding of this is `bosh import [toml | yaml | json | cl_template] [config file]` will create a descriptor, similar to how `bosh import docopt [file]` creates a base descriptor.
The output descriptor will contain `inputs`, and `output-files` (contains a config_file with `file-template`). Is this correct?

Hi @glatard, does it mean `bosh create [output_desc] --cl-template [CL_template]` will populate `inputs` of the `[output_desc]` according to the command-line template?

Thanks for the clarifications!

gkiar:
@DarrinFong exactly on my point, and I'm pretty sure you're also spot on for @glatard 's :)

DarrinFong:
@gkiar I'm having trouble finding config files to test with, do you have some config_file/descriptor that I can validate the generated descriptor against? In case the descriptor for a config file doesn't exist, I can still use the config file to test the import and manually verify the output descriptor. Thanks :)

PR #611, opened 2020-05-12
Bimport estimate descriptor with config file

DarrinFong:
## Related issues

#575 

## Purpose

Using config-file with default-values, generate base descriptor with `command-line`, `inputs` and an `output-files` item: `config_file`.
Using a descriptor containing a config-file output (with file-template), generate base descriptor with `inputs`.

## Current behaviour

N/A

## New behaviour

`bosh import config [output_descriptor] [config_file.{json|toml|yml}]` will create a base descriptor with:
- command-line (tool name placeholder, inputs, and config_file)
- inputs (estimated properties: id, name, type, integer, list, default-value, value-key")
- config_file (path-template based on config_file extension, file-template)

`bosh create [output_descriptor] --cl-template [template_descriptor]` will create a base descriptor with:
- inputs (All properties are based on value-key: id, name, type: String, value-key)
 
Examples: 
`bosh import config json_output.json boutiques/tests/config/configuration.json`
`bosh import config toml_output.json boutiques/tests/config/configuration.toml`
`bosh import config yaml_output.json boutiques/tests/config/configuration.yml`
`bosh create json_config_inputs.json --cl-template boutiques/tests/config/json_config_desc.json`
`bosh create toml_config_inputs.json --cl-template boutiques/tests/config/toml_config_desc.json`
`bosh create yaml_config_inputs.json --cl-template boutiques/tests/config/yml_config_desc.json`

## Implementation Detail

`bosh import`:
- config_file import only works on valid extensions {.json|.toml|.yml} for import validation purposes (imported using python libraries)
- inputs are imported by key:value pair, input properties are estimated from value
- file-template is created using imported inputs, formatting depends on config_file extension

`bosh create`:
- inputs are generated from file-template, no meaningful properties can be estimated using only value-key

Note: [oyaml](https://github.com/wimglenn/oyaml) replaced pyyaml so loading preserves ordering.


DarrinFong:
Hi @gkiar, can you review this PR and let me know if the configuration files and descriptors in `tools/python/boutiques/tests/config/` are representative of config files used by some tools? Thanks a lot!

gkiar:
Hey @DarrinFong this looks great!! One question I'd have is how the tool handles indented values (i.e. just a nested structure in JSON, or something [like this](https://github.com/boutiques/boutiques/blob/master/.travis.yml#L24-L27) for yaml).

A few other comments:
- The arguments contained in the config file often don't also appear on the command line, but in the first example here, for instance, you have them both on the command-line template and in the config file. I think we only want the latter.
- Could you just clarify the point of the `create` commands to me? I'm not quite sure I understand what is being generated or what the use case is.

This is really great, nice work :)

DarrinFong:
@gkiar Indented values are not supported at the moment because I don't know how to map them to a descriptor, I can add support if there's an example config file.
- I removed the inputs from cml
- I totally misunderstood the purpose of `create`, I made some changes and this is the new behaviour: `bosh create output_desc --cl-template input_desc` will create a template `output_desc` with inputs based on the `input_desc`'s command-line (or a string).

DarrinFong:
Hey @gkiar, do you think this issue is ready to be merged? Thanks!

Issue #630, opened 2020-07-21
The tool's "shell" is appended with "-l" in the execution script


prioux:
A boutiques descriptor can have an optional "shell" specified other than the default "/bin/sh". The value of that shell attribute is used to build the first line of a local executions script (a `#!`
 script) when runnning in a singularity image.

However, boutiques always append " -l" (space dash ell) to the shell's path in that script, which won't work when the shell invoked doesn't have a `-l` option.

E.g. if the descriptor has:

```json
{
  "name": "test",
  "tool-version": "0.0.1",
  "schema-version": "0.5",
  "description": "Error Report",
  "author": "Pierre Rioux",
  "command-line": "print('Hi');",
  "shell": "/usr/bin/python3",
   "etc": "not shown"
}
```

bosh will attempt to create a script that looks like

```python
#!/usr/bin/python3 -l
print('Hi');
```

And the execution will fail immediately with:

```
Unknown option: -l
usage: /usr/bin/python3 [option] ... [-c cmd | -m mod | file | -] [arg] ...
Try `python -h' for more information.
```

## Expected Behavior
Not to assume all shells have a "-l" option.

## Context (optional)
Only happens when the commad is to be run in a container.

prioux:
My suggestion: don't even attempt to override what the descriptor says. Just put the value of "shell" directly after the "#!".

PR #635, opened 2020-08-18
-l flag only applied if shell path is /bin/sh

DarrinFong:
## Related issues

#630 

## Purpose

To not append `-l` when calling a custom shell.

## Current behaviour

Boutiques always append " -l" (space dash ell) to the shell's path in that script, which won't work when the shell invoked doesn't have a -l option.

## New behaviour

`-l` is only appended to script when calling the default `/bin/sh`


glatard:
For backward compatibility, `-l` is kept only for the default shell, `/bin/sh`.

Issue #644, opened 2021-05-25
Disables-inputs is valid between non-optional inputs


draguar:
## Expected Behavior

It is my understanding that 'disables-inputs' property allows one to specify that if current input is provided, listed inputs can't be. If this is correct, this should only be valid if the inputs to disable are also optional. 
## Current Behavior

A descriptor containing an optional input listing non-optional inputs in its 'disables-inputs' array passes bosh validate.

## Context (optional)

I am trying to automatically create web client forms from applications' boutiques descriptor. For this, I assume that provided descriptors are valid, which I check with bosh validate. 
## Possible fix (optional)

If my expected behavior indeed is the intended behavior, bosh validate simply needs to check if listed inputs are optional.

PR #646, opened 2021-06-18
created minimal install bundle

### LINKED ISSUE #645: Have boutiques[minimal] install
Boutiques has a growing list of dependencies, and for the basic use case we just want the ability to validate and run a tool. If we use [`setuptools[Extras]`](https://packaging.python.org/tutorials/installing-packages/#id29), we could have a `boutiques[all]` which still gives us everything, but the default would be `boutiques`/`boutiques[minimal]`, which only has this core functionality. Reducing unnecessary dependency clutter, including things like clients for the Zenodo API in cases where execution records and descriptors are managed locally, would lower the barrier to adoption in other platforms.

gkiar:
## Checklist

- [x] **DO** Unit tests pass.

## Purpose

Makes Boutiques lighter

## Current behaviour

Installation with `pip install boutiques` would install every dependency, even if I didn't plan on using many functions w/in the lib.

## New behaviour

Now, the base install comes with the minimal set of dependencies to use/validate descriptors, and all the rest can be installed with `pip install boutiques[all]`.
 
- [x] No

## Implementation Detail
- added `importCatcher()` decorator to wrap functions which require non-essential dependencies
- moved non-essential dependency imports to within functions using them
- the above closes #645 
- updated docs
- also addressed #644 by adding a validation condition

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ install:
 - pip install --upgrade setuptools
 - pip install "jsonschema>=2.6.0,<3.0.0"
 - pip install pytest pytest-runner pycodestyle coveralls
-- pip install .
+- pip install .[all]
 - pip freeze
 
 script:
--- a/README.md
+++ b/README.md
@@ -25,8 +25,11 @@ Simple! Just open your favourite terminal and type:
 
     $ pip install boutiques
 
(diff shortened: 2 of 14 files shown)


glatard:
Thanks for doing that @gkiar!

PR #656, opened 2023-09-28
Allow missing tool version, assigning "unknown"

### LINKED ISSUE #655: Verbose search fails when expected metadata is missing
TLDR: `bosh search -v` errors out right now.

When doing a verbose search, boutiques assumes every Zenodo record it finds will have a version specified under its metadata. If this is not the case, processing the search JSON will fail with a `KeyError`:

https://github.com/boutiques/boutiques/blob/5a2374e87dd837cb83fec9603f32e3c9085613ba/tools/python/boutiques/searcher.py#L121

In the last week or so, a couple of Boutiques tools have been published ([This one](https://zenodo.org/record/8357249) and [this one](https://zenodo.org/record/8365022), check the [JSON](https://zenodo.org/record/8357249/export/json) [export](https://zenodo.org/record/8365022/export/json) to confirm) that have no version defined in their metadata, causing a verbose search for all Boutiques tools to fail without returning anything.

Instead of failing, I think Boutiques should either:

1. Not require a version from tools and process them without.
2. Skip tools that don't have a version, possibly raising a warning.

I can go ahead and make either of these changes in a PR depending on what you think

tkkuehn:
## Related issues
Fixes #655 

## Checklist

- [x] **DO** Unit tests pass.

## Purpose
This allows Zenodo records without a version specified to be processed.

## Current behaviour
Bosh search throws a KeyError before returning anything.

## New behaviour
Boutiques will return a record with "VERSION" set to "unknown".
 
- [ ] Yes
- [x] No

### DIFF
--- a/tools/python/boutiques/searcher.py
+++ b/tools/python/boutiques/searcher.py
@@ -118,7 +118,7 @@ class Searcher():
         for hit in results["hits"]["hits"]:
             (id, title, description, downloads) = self.parse_basic_info(hit)
             author = hit["metadata"]["creators"][0]["name"]
-            version = hit["metadata"]["version"]
+            version = hit["metadata"].get("version", "unknown")
             publication_date = hit["metadata"]['publication_date']
             doi = hit["doi"]
             keyword_data = self.get_keyword_data(hit["metadata"]["keywords"])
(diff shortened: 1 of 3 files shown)


glatard:
Thanks a lot @tkkuehn, this will be part of the upcoming release.

PR #657, opened 2023-10-03
Use != instead of is not with literal

yarikoptic:
addresses warnings I got

    /home/yoh/proj/misc/boutiques/tools/python/boutiques/localExec.py:1066: SyntaxWarning: "is not" with a literal. Did you mean "!="?
      if val is not None and val is not "":
    /home/yoh/proj/misc/boutiques/tools/python/boutiques/prettyprint.py:125: SyntaxWarning: "is not" with a literal. Did you mean "!="?
      if config_info is not "Config Files:" else ""
    /home/yoh/proj/misc/boutiques/tools/python/boutiques/prettyprint.py:318: SyntaxWarning: "is not" with a literal. Did you mean "!="?
      if req_inp_descr is not "":
    /home/yoh/proj/misc/boutiques/tools/python/boutiques/prettyprint.py:323: SyntaxWarning: "is not" with a literal. Did you mean "!="?
      if opt_inp_descr is not "":

- [ ] Yes
- [x] No

### DIFF
--- a/tools/python/boutiques/localExec.py
+++ b/tools/python/boutiques/localExec.py
@@ -1063,7 +1063,7 @@ class LocalExecutor(object):
                        template.find(clk) > 0 and is_output):
                         val = op.basename(val)
                 # Here val can be a number so we need to cast it
-                if val is not None and val is not "":
+                if val is not None and val != "":
                     template = template.replace(clk, str(val))
                 else:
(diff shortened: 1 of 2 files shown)


glatard:
Thanks @yarikoptic, this will be part of the upcoming release.

Issue #650, opened 2023-03-20
Support for apptainer (upgrade/migrate from singularity)


sandepat:
## Purpose
Apptainer (https://apptainer.org/), which is the Linux Foundation variant of Singularity, to be supported by boutiques

## Context
Singularity has moved to apptainer (for now they are identical apart from the name) and the system admins have started upgrading systems to support apptainer. It could be interesting for boutiques to integrate support to apptainer, as it currently doesn't seem to have one. 

## Possible Implementation
since all the commands of singularity are identical to that of apptainer (as of now), the integration would be to simply migrate from singularity to apptainer without requiring several additional changes (while ideally still supporting singularity as well)  - (https://apptainer.org/docs/admin/main/singularity_migration.html)

mathdugre:
Resolved by https://github.com/boutiques/boutiques/pull/707

PR #688, opened 2024-11-07
Support Apptainer

sandepat:
---

name: Apptainer Support Integration  
about: Propose modifications to add Apptainer container support alongside existing Singularity and Boutiques support.  
title: 'Feature implementation | Apptainer support integration'  
labels: enhancement  
assignees: ''  

---

## Related issues

No specific issue.

## Checklist

- [x] **DO** Unit tests pass.
- [x] **DO** Created unit tests for Apptainer support.
- [x] **TRY** Documented Apptainer-specific additions (not applicable for an API function).

## Purpose

This PR introduces support for Apptainer (formerly Singularity) within the system, expanding container compatibility and improving workflow flexibility for containerized applications.

## Current behaviour

Currently, the system supports container execution using Docker and Singularity only.

## New behaviour

With Apptainer support integrated, users can now leverage Apptainer containers in addition to existing Docker and Singularity options. This ensures compatibility with diverse computing environments and enhances reproducibility and portability.

- [ ] Yes
- [x] No

## Implementation Detail

This update includes:
- Configurations to support Apptainer environment variables.
- Functions to pull and execute Apptainer containers, similar to Singularity handling.
- Tests to validate Apptainer functionality and ensure smooth integration with the existing workflow

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -18,7 +18,7 @@ assignees: ''
  the information needed to reproduce the bug (descriptors, inputs, etc) -->
 
 ## Environment (optional)
-<!--- Information about the execution environment, in particular: Python version, Docker/Singularity version
+<!--- Information about the execution environment, in particular: Python version, Docker/Singularity/Apptainer version
  if relevant -->
 
 ## Context (optional)
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,16 @@ before_install:
 - docker version # Docker installation check
 - singularity version # Singularity installation check
 
+# Install Apptainer
+- wget https://github.com/apptainer/apptainer/releases/download/v1.1.7/apptainer-1.1.7.tar.gz
+- tar xzf apptainer-1.1.7.tar.gz
+- cd apptainer-1.1.7
+- ./mconfig
+- make -C ./builddir
+- sudo make -C ./builddir install
+- cd ${TRAVIS_BUILD_DIR}/
+- apptainer version # Apptainer installation check
+
 install:
 - pip install --upgrade setuptools
 - pip install "jsonschema>=2.6.0,<3.0.0"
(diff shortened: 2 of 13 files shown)


glatard:
Hi @sandepat, thank you for opening this PR. Since apptainer is really a dropin replacement for Singularity, shall we just replace "Singularity" with "Apptainer" everywhere instead of supporting both option names? For instance, `--use-singularity` would become `--use-apptainer` but we wouldn't support both. To preserve backward compatibility, only the schema would need to support both options. What do you think?

@mathdugre, any opinion?

ngeorges-cnrs:
Hi @glatard @mathdugre, I'm joining this PR as @sandepat recently left the team at https://github.com/virtual-imaging-platform, who was at the origin of this request.

We only recently clarified the history and current status of the Singularity project, with its 2021 fork into [Apptainer](https://github.com/apptainer/apptainer) on one hand and [Singularity-CE](https://github.com/sylabs/singularity) on the other hand, which sheds some new light on this issue. As the historical Singularity project is now obsolete, we think that it would indeed make more sense to just replace "Singularity" with "Apptainer" everywhere, and keep both terms only when strictly needed for backwards compatibility. Supporting "singularity" in addition to "apptainer" in the future would more or less imply that boutiques also supports singularity-CE, which has kept the `singularity` command name. This project already has a number of [differences](https://github.com/sylabs/singularity/discussions/2948) with apptainer, and those will likely only grow and increase the maintenance cost in boutiques.

If we confirm the intent of just replacing "Singularity" with "Apptainer", including references to https://singularity.lbl.gov/ with https://apptainer.org/, let me know, and I can then update this PR accordingly.

glatard:
Thank you @ngeorges-cnrs, for the detailed explanation. Since apptainer still provides the `singularity` command, what would be the limitation to just support `singularity` then, and explain in the documentation that apptainer would be called through the `singularity` command? We would then support both apptainer and singularityCE with minimal changes.

ngeorges-cnrs:
@glatard : I agree this would also be a valid option. In my opinion, it's "just" a matter of clarifying which container runtimes boutiques officially supports, now that "singularity" has become ambiguous :
- if both apptainer and singularity-ce are supported, then make that official in the doc, including the fact that both are called through the same `singularity` command, and update deprecated references to https://singularity.lbl.gov/ to replace them with the new websites of these two projects. In that case, there are probably no code changes required in boutiques, only doc updates, but there might be some extra maintenance cost down the road to maintain compatibility with both runtimes, depending on how much they diverge. It would still be possible to drop support of one of them at some point if that's relevant.
- or, if the decision is to support only apptainer, then I'd favor replacing all references to singularity with apptainer to avoid confusion with singularity-ce, and keeping only the minimal things for backwards compatiblity.

camarasu:
Discussions with @glatard and @ngeorges-cnrs led to the conclusion that this PR can be closed. 
Another solution will be proposed in reply to issue #650

PR #707, opened 2025-03-07
Feature implementation for issue #650 - apptainer support

ngeorges-cnrs:
## Related issues
#650, #688

## Checklist
- [x] **DO** Unit tests pass.
- [x] Documentation: no new API function, one new command-line flag documented in `bosh exec launch --help`

- [ ] Yes
- [x] No

## Purpose
Add support for apptainer runtime as discussed in [#688](https://github.com/boutiques/boutiques/pull/688#issuecomment-2612600351) and [#650](https://github.com/boutiques/boutiques/issues/650#issuecomment-2612633629).

All changes in this PR should be backwards-compatible, the intent is to support hypothetical environments where the `apptainer` command exists but `singularity` doesn't, and to allow explicit selection of the `apptainer` runtime with `bosh exec launch --force-apptainer`.

Detail of changes:
- possible values for the `container-image.type` descriptor field remain unchanged: only `docker` or `singularity`
- when the `singularity` command is not found, but the `apptainer` command exists, then call apptainer instead of failing. This is handled in `localExec.py:_chooseContainerTypeToUse()`
- add a `--force-apptainer` flag on `bosh exec launch`, which makes the apptainer command take precedence over singularity, in a similar way to the existing flags `--force-singularity` and `--force-docker`
- replace deprecated link to https://singularity.lbl.gov/ with https://apptainer.org/ in `GETTING_STARTED.md`

Some points of attention:
- this PR is based on `develop` as recommended in the contributing guidelines, but the master branch is currently more recent than develop. There is no conflict with either branch at the time of this PR.
- the tests added for this feature were marked as XFAIL, as they are very similar to tests which were also marked XFAIL a few months ago in https://github.com/boutiques/boutiques/commit/fd81dddfedf1edfb3e8f250677bc8c22e1836642. They do work on a local machine.
- I had some hesitation on what to do in the edge case where a `--force-X` flag is used and command X is not installed. For now, I favored backwards compatibility and kept the existing behaviour of defaulting to other available commands. This can already be observed in the existing boutiques implementation: when `--force-docker` is used and `docker` is not installed but `singularity` is, then bosh uses singularity. In the same vein, this PR can end up running `apptainer` even `--force-singularity` is used. If this is not desired behaviour, then we could modify `_chooseContainerTypeToUse()` to do an early abort when `--force-X` is used on an unavailable command.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -10,7 +10,7 @@ This guide will help you set up your local environment for coding and testing Bo
 -   **pip**. Most versions of Python will come with it, but in case you don't have it, instructions to install it are [here](https://pip.pypa.io/en/stable/installing/).
 -   **Docker or Singularity**. Depending on the tools you plan to work with and where they are installed, you'll want to have either Docker or Singularity installed (ideally both).
   -   To install Docker, follow the  [instructions](https://docs.docker.com/install/overview/) to install Docker CE.
-  -   To install Singularity, follow the [instructions](https://singularity.lbl.gov/install-linux) to download the latest stable release of Singularity.
+  -   To install Singularity, follow the [instructions](https://apptainer.org/docs/admin/latest/installation.html#installation-on-linux) to download the latest stable release of Singularity.
 -   **Root (sudo) access**. If your system doesn't already have all of the above installed, you'll need root permission to install new programs.
 
 ## 1.   Clone the boutiques repository
--- a/boutiques/bosh.py
+++ b/boutiques/bosh.py
@@ -117,6 +117,7 @@ def execute(*params):
                 "skipDataCollect": results.skip_data_collection,
                 "forceDocker": results.force_docker,
                 "forceSingularity": results.force_singularity,
+                "forceApptainer": results.force_apptainer,
                 "provenance": results.provenance,
                 "noContainer": results.no_container,
                 "sandbox": results.sandbox,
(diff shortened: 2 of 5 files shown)


ngeorges-cnrs:
Hi @mathdugre,

I saw the failing CI on two github actions:
- for the style check https://github.com/boutiques/boutiques/actions/runs/13724159763/: my bad, I didn't use tox correctly before submitting this PR. I've fixed the style in a new commit https://github.com/boutiques/boutiques/pull/707/commits/76635545de63df3bce9839abe78794ec5bea9de1, so this should pass now.
- for the 10 unit tests jobs in https://github.com/boutiques/boutiques/actions/runs/13724159760/, the error is always the same:
```
FAILED boutiques/tests/test_data_handler.py::TestDataHandler::test_publish_nexus_invalid_token - AssertionError: 'Cannot authenticate to Nexus API, check your access token' not found in '[ ERROR ] Cannot access Nexus endpoint'
```
As far as I understand, this is an existing problem with boutiques CI, and unrelated to the present PR. Just let me know if I'm mistaken, or if I can help in any way.

mathdugre:
Hi @ngeorges-cnrs, sorry for the delayed reply.

Since `develop` branch was merged in `master`; I'll modify this PR to target `master` as well.
#704 addressed the issue with NEXUS test failing. Could you update your branch with `master`, this should fix the issue.

ngeorges-cnrs:
@mathdugre : thanks for your response and clarification on the target branch.
I've updated my branch with `master`, all tests should pass now.

PR #713, opened 2025-07-17
Omit test folders and files from coverage.

mathdugre:
---
name: Feature or bug fix name
about: Propose modifications in the code
title: 'Feature implementation | Bug fix for issue #<000>'
labels: enhancement
assignees: ''

---

## Checklist

- [ ] **DO** Unit tests pass.

## Current behaviour

Currently codecov runs on the test folders and files. This prevents proper reporting of the coverage.

- [ ] Yes
- [ ] No

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,6 +68,13 @@ line-length = 88
 ignore-words-list = "nd,fo,minum,manum,assertIn"
 skip = ".git,*.pdf,*.svg,*.json"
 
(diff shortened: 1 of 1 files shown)


mathdugre:
@michellewang, could you review this?

michellewang (approved):
Looks good to me, there are less misses in the Codecov report (though also less hits, so total coverage is lower...), so this seems to be doing something

Issue #714, opened 2025-07-18
Behavior of --force-docker


ngeorges-cnrs:
Hello,

Following the recent merge of #707 (thanks!), my team and I would like to discuss a remaining point and propose an extra change regarding how `bosh exec launch` behaves with `--force-X` flags. This point was also mentioned among secondary concerns in #710, but I'll recapitulate everything here for clarity.
It all boils down to how `_chooseContainerTypeToUse()` and `_isCommandInstalled()` are implemented in `localExec.py` :

### Current status

- Behaviour in versions <=0.5.30 :
  - `--force-docker` : run docker if command is found in `$PATH`, and look for `singularity` otherwise. So in fact this is more like "trying docker" rather than really "forcing docker". The `--help` documentation uses the word `tries`, but the flag is named `force`.
  - `--force-singularity` : run singularity if command is found in `$PATH`, fail otherwise

- Behaviour on master branch since the merge of #707 :
  - `--force-docker` : same as above
  - `--force-singularity` : run singularity if command is found in `$PATH`, and look for `apptainer` otherwise
  - `--force-apptainer` : run apptainer if command is found in `$PATH`, fail otherwise

### Proposed change

For all three `--force-X` flags on `bosh exec launch`, run `X` if command is found in `$PATH`, and immediately fail with `"Could not find container engine X"` if command is not found. Note that this would happen _only_ when a `--force` option is used : if no such option is used then the docker/singularity/apptainer fallback would still occur.

Besides being arguably "less surprising" from a user perspective given the "force" terminology, this change would also allow `--container-opts` to be usable with `--force-docker` on a `"type":"singularity"` descriptor, which is currently not possible as stated in #710.

---

Please let me know if such a change sounds relevant and low-risk enough to the boutiques maintainers : in case it is, then I can work on a PR later in August/September.

mathdugre:
Hi @ngeorges-cnrs , thank you for reiterating and detailing the issue.
I'll discuss this with the other maintainers and follow up here.

mathdugre:
Hi @ngeorges-cnrs, I discussed the proposed solution with @glatard and we agree that this would be an improvement, despite a minor break in backward compatibility.

We would be happy to welcome a PR.

---
@michellewang (FYI)

PR #715, opened 2025-08-21
Update --force-X flags

ngeorges-cnrs:
## Related issues
#714, #710

## Checklist
- [x] **DO** Unit tests pass.
- [x] Documentation: no change.

- [ ] Yes
- [x] No. But there is a small breakage of backward compatibility, as discussed in issue #714.

## Purpose

Change how `--force-docker`,`--force-singularity`,`--force-apptainer` flags are processed on `bosh exec launch`, so that either the forced command is found in `$PATH` and used, or an error is raised.
This small behaviour change fixes an arguable situation where a command other than the "forced" one could end up being run, and also allows `--container-opts` to now be usable with `--force-docker` (addressing the limitation mentionned in #710).
Behaviour when no force flag is used is unchanged.

## Current behaviour

- `--force-docker` : run docker if command is found in `$PATH`, and look for `singularity` then `apptainer` otherwise
- `--force-singularity` : run singularity if command is found in `$PATH`, and look for `apptainer` otherwise
- `--force-apptainer` : run apptainer if command is found in `$PATH`, fail otherwise

## New behaviour

- `--force-docker` : run docker if command is found in `$PATH`, fail otherwise
- `--force-singularity` : run singularity if command is found in `$PATH`, fail otherwise
- `--force-apptainer` : run apptainer if command is found in `$PATH`, fail otherwise

### DIFF
--- a/boutiques/localExec.py
+++ b/boutiques/localExec.py
@@ -246,9 +246,9 @@ class LocalExecutor:
             (con.get("image") if not self.noContainer else None),
         )
         conIndex = con.get("index")
-        # conOpts can't be forced for docker, as long as there are cases
-        # where _chooseContainerTypeToUse() can fall back to singularity
-        conOptsForced = conOpts is not None and self.forceSingularity
+        # always preserve conOpts when container command is forced
+        conForcedCommand = self._getContainerForcedCommand()
+        conOptsForced = conOpts is not None and conForcedCommand is not None
         conOpts = conOpts or con.get("container-opts")
         conIsPresent = conImage is not None
         # Export environment variables,
@@ -282,7 +282,7 @@ class LocalExecutor:
         if conIsPresent:
             # Figure out which container type to use
             (conTypeToUse, conBinName) = self._chooseContainerTypeToUse(
-                conType, self.forceSingularity, self.forceApptainer, self.forceDocker
+                conType, conForcedCommand
             )
             # Pull the container
             (conPath, container_location) = self.prepare(conTypeToUse, conBinName)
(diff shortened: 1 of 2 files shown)


ngeorges-cnrs:
Hi @mathdugre,

A few words on CI errors :
- The failing "Test with 3.10 on ubuntu-latest" seems to be a temporary network failure with Zenodo, so I'm assuming a false positive here.
> FAILED boutiques/tests/test_pull.py::TestPull::test_pull_missing_raises_exception - AssertionError: 'Descriptor "9999990" not found' not found in '[ ERROR (504) ] Error searching Zenodo'
- The [2 missing lines](https://app.codecov.io/gh/boutiques/boutiques/pull/715?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=boutiques) in codecov are related to the case where bosh ends up running apptainer when no force option is specified. This scenario was in fact already not tested in #650, but this was not detected by codecov with the previous implementation, because this condition was on the same line as the forceApptainer flag. We can cover these two lines by adding an extra unit test similar to `test_example1_exec_docker_not_installed`, with a mock function similar to `docker_not_installed` that would report both docker and singularity as False, thus causing apptainer to be selected.
Just let me know if this extra test is worth adding, and also whether it should be marked xfail just as `test_example1_exec_docker_not_installed` (I guess it should be xfail, but I'm still not really sure why).

mathdugre (approved):

mathdugre:
LGTM 🎉

PR #728, opened 2025-12-09
Feature/styx support

astewartau:
---
name: Add Styx schema extensions support
about: Propose modifications in the code
title: 'Feature implementation: Add support for Styx (0.5+styx) schema extensions'
labels: enhancement
assignees: ''

---

## Related issues

- Enables Boutiques to validate and execute descriptors from [NiWrap](https://github.com/childmindresearch/niwrap), which provides ~1,700 neuroimaging tool wrappers using the Styx schemaformat.

## Checklist

- [x] **DO** Unit tests pass.

## Purpose

Add support for the Styx schema extensions (schema-version `0.5+styx`), which extends Boutiques 0.5 with additional features for type-safe code generation while maintaining backward compatibility.

## Current behaviour

- `bosh validate` rejects descriptors with `schema-version: "0.5+styx"`
- `bosh validate` rejects descriptors with nested input types (where `type` is an object instead of a string)
- `bosh validate` rejects descriptors with `stdout-output` or `resolve-parent` fields
- `bosh exec launch` corrupts Flag inputs when input/output share the same ID
- `bosh exec launch` fails on tools with non-UTF8 output

## New behaviour

- `bosh validate` accepts `schema-version: "0.5+styx"`
- `bosh validate` accepts nested input types, `stdout-output`, and `resolve-parent` fields
- `bosh exec launch` correctly handles Flag inputs with duplicate IDs
- `bosh exec launch` handles non-UTF8 output gracefully

- [x] Yes
- [ ] No

### 1. Schema Changes (`boutiques/schema/descriptor.schema.json`)

| Change | Description |
|--------|-------------|
| `schema-version` | Added `"0.5+styx"` as valid enum value |
| Input `type` | Changed from string enum to `oneOf` supporting both primitive types (`String`, `File`, `Flag`, `Number`) and nested type objects |
| `stdout-output` | New top-level property for tools returning structured stdout |
| `resolve-parent` | New input property (boolean) for parent reference resolution |

**Nested input types** allow describing sub-command structures:
```json
{
    "id": "config",
    "type": {
        "id": "config",
        "name": "config",
        "command-line": "-config [KEY] [VALUE]",
        "inputs": [...],
        "output-files": []
    },
    "list": true
}
```

### 2. Validator Changes (`boutiques/validator.py`)

- Updated `schema_types` extraction to handle the new `oneOf` structure
- Added `getInputTypeName()` helper to extract type from both string and nested object types

### 3. Execution Fixes (`boutiques/localExec.py`)

- **Flag handling fix**: Skip `escape_string()` for boolean values to prevent corruption
- **Duplicate ID fix**: Use `in_dict` instead of merged `in_out_dict` for Flag inputs when input/output share IDs
- **Unicode fix**: Added `errors='replace'` to `decode()` for tools with non-UTF8 output

### 4. Relaxed Validations (Separate Commit)

The following validations were relaxed for Styx compatibility:

| Relaxation | Rationale | NiWrap Usage |
|------------|-----------|--------------|
| `tool-version` optional | Styx uses version from directory structure | All tools |
| `output-files` can be empty | Some tools only have stdout output | ~143 tools |
| IDs unique per category (not globally) | Allows input `binary_mask` + output `binary_mask` | ~30 tools |
| Flags don't require `optional: true` | Styx marks all flags as required in schema but optional in practice | All tools |

**Discussion**: Should these relaxations be:
- A) Accepted as-is (more permissive for all descriptors)
- B) Conditional on `schema-version: "0.5+styx"` only
- C) Rejected (require Styx/NiWrap to fix their descriptors)

### Testing

- Validated against NiWrap's FSL bet, MRTrix dwidenoise, and AFNI 3dCM descriptors
- `bosh validate` passes for Styx descriptors
- `bosh exec launch` correctly handles Flag inputs with `false` values

### Commits

1. **Add Styx schema extensions support** - Core schema changes
2. **Relax validation for Styx compatibility** - Permissive validation changes
3. **Fix Flag input handling for duplicate IDs and unicode errors** - Execution fixes

### DIFF
--- a/boutiques/schema/descriptor.schema.json
+++ b/boutiques/schema/descriptor.schema.json
@@ -145,7 +145,7 @@
             "id": "http://github.com/boutiques/boutiques-schema/schema-version",
             "type": "string",
             "description": "Version of the schema used.",
-            "enum": ["0.5"]
+            "enum": ["0.5", "0.5+styx"]
         },
         "environment-variables": {
             "id": "http://github.com/boutiques/boutiques-schema/environment-variables",
@@ -266,9 +266,42 @@
                     },
                     "type": {
                         "id": "http://github.com/boutiques/boutiques-schema/input/type",
-                        "type": "string",
-                        "description": "Input type.",
-                        "enum": ["String", "File", "Flag", "Number"]
+                        "description": "Input type. Can be a primitive type string or a nested type object (Styx extension).",
+                        "oneOf": [
+                            {
+                                "type": "string",
+                                "enum": ["String", "File", "Flag", "Number"]
+                            },
+                            {
+                                "type": "object",
+                                "description": "A nested input type that defines a sub-command structure (Styx extension).",
+                                "properties": {
+                                    "id": {
+                                        "type": "string",
+                                        "pattern": "^[0-9,_,a-z,A-Z]*$",
+                                        "minLength": 1
+                                    },
+                                    "name": {
+                                        "type": "string",
+                                        "minLength": 1
+                                    },
+                                    "description": {
+                                        "type": "string"
+                                    },
+                                    "command-line": {
+                                        "type": "string",
+                                        "minLength": 1
+                                    },
+                                    "inputs": {
+                                        "type": "array"
+                                    },
+                                    "output-files": {
+                                        "type": "array"
+                                    }
+                                },
+                                "required": ["id", "name", "command-line", "inputs", "output-files"]
+                            }
+                        ]
                     },
                     "description": {
                         "id": "http://github.com/boutiques/boutiques-schema/input/description",
@@ -389,6 +422,11 @@
                         "id": "http://github.com/boutiques/boutiques-schema/input/uses-absolute-path",
                         "description": "Specifies that this input must be given as an absolute path. Only specifiable for File type inputs.",
                         "type": "boolean"
+                    },
+                    "resolve-parent": {
+                        "id": "http://github.com/boutiques/boutiques-schema/input/resolve-parent",
+                        "description": "Whether to resolve parent input references (Styx extension).",
+                        "type": "boolean"
                     }
                 },
                 "required": [
@@ -404,6 +442,10 @@
                 }, {
                     "properties": {
                         "type": { "enum": ["String", "File", "Number"] }
+                    }
+                }, {
+                    "properties": {
+                        "type": { "type": "object" }
                     }
                 }],
                 "dependencies": {
@@ -737,6 +779,29 @@
         "minItems": 1,
         "uniqueItems": true
     },
+        "stdout-output": {
+            "id": "http://github.com/boutiques/boutiques-schema/stdout-output",
+            "type": "object",
+            "description": "Structured output captured from command stdout (Styx extension).",
+            "properties": {
+                "id": {
+                    "type": "string",
+                    "pattern": "^[0-9,_,a-z,A-Z]*$",
+                    "minLength": 1,
+                    "description": "Unique identifier for the stdout output."
+                },
+                "name": {
+                    "type": "string",
+                    "minLength": 1,
+                    "description": "Human-readable name for the stdout output."
+                },
+                "description": {
+                    "type": "string",
+                    "description": "Description of the stdout output."
+                }
+            },
+            "required": ["id"]
+        },
         "custom": {
             "id": "http://github.com/boutiques/boutiques-schema/custom",
             "type": "object"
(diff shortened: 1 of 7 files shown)


michellewang:
Putting it here so that I can find the link later: the Styx docs summarize the changes made to the 0.5 schema https://niwrap.dev/styxbook/boutiques_guide/index.html

michellewang (commented):
Hi @astewartau 

Thanks for opening the PR (and sorry for the long delay)!

It would be great to get the Styx extensions back into Boutiques proper. I took an initial look, and in general the changes to the schema look good to me. I think we'll need to take a deeper look at the changes in `localExec.py` and `validator.py` as I am not really familiar with that part of the codebase.

Some initial thoughts (also pinging @mathdugre who should take a look as well):
1. Would it make more sense to have a dedicated schema file for `0.5+styx`? Otherwise, a descriptor could theoretically have Styx extensions with the old `schema-version` 0.5 but still be valid, which feels wrong.
2. Would it be possible to reduce the redundancy in the `type` list? Specifically between the `array` and `object` options as `array` is just wrapping `object`, so they should always be in sync.

I also did a bit of manual testing with some Nipoppy pipelines as a sanity check, and so far there is nothing that broke.

boutiques/tests/test_styx.py
```diff
@@ -0,0 +1,62 @@
...
+        """Test that stdout-output field is accepted."""
+        # The example_styx.json has a stdout-output field
+        fil = self.get_file_path("example_styx.json")
+        self.assertIsNone(bosh(["validate", fil]))
+
+    def test_styx_resolve_parent(self):
+        """Test that resolve-parent field on inputs is accepted."""
+        # The example_styx.json has resolve-parent: true on topup_result input
+        fil = self.get_file_path("example_styx.json")
+        self.assertIsNone(bosh(["validate", fil]))
```
mathdugre, line 43: These tests seem to be duplicates. I would suggest to either
1. Combine them together, potentially with the other ones (`test_styx_nested_type_conditional_path_template` & `test_styx_union_type`)
2. Create individual descriptors to test the new functionalities.

In either cases, we could parametrize the tests over the descriptor name.

boutiques/schema/descriptor.schema.json
```diff
@@ -145,7 +145,7 @@
             "id": "http://github.com/boutiques/boutiques-schema/schema-version",
             "type": "string",
             "description": "Version of the schema used.",
-            "enum": ["0.5"]
+            "enum": ["0.5", "0.5+styx"]
```
mathdugre, line 148: I'm uncertain of calling the new schema `0.5+styx`. It might be worth considering a `0.6` schema version.
@glatard @gkiar , what do you think?

michellewang:
Ran into an issue today with when testing a descriptor generated with [`argdump`](https://github.com/styx-api/argdump) and [`styx-ts`](https://github.com/styx-api/styx-ts), where `bosh validate` passes but `bosh exec simulate` fails: looks like the invocation schema cannot be generated correctly from a descriptor with nested input types.

The error can be replicated with the [`example_styx.json`](https://github.com/astewartau/boutiques/blob/9189551d4f46f4abfee4f671775930767570dc07/boutiques/tests/styx/example_styx.json) file and a minimal invocation:

```
{
    "input_file": "/my/input/file",
    "output_name": "my_output_name"
}
```

Exception traceback:

```
  File "/data/origami/michelle/projects/boutiques/.venv/bin/bosh", line 10, in <module>
    sys.exit(bosh())
             ~~~~^^
  File "/data/origami/michelle/projects/boutiques/boutiques/bosh.py", line 542, in bosh
    out = execute(*params)
  File "/data/origami/michelle/projects/boutiques/boutiques/bosh.py", line 143, in execute
    valid = invocation(*arguments)
  File "/data/origami/michelle/projects/boutiques/boutiques/bosh.py", line 308, in invocation
    invSchema = generateInvocationSchema(descriptor)
  File "/data/origami/michelle/projects/boutiques/boutiques/invocationSchemaHandler.py", line 80, in generateInvocationSchema
    schema["properties"] = reduce(addTypeConstraints, inputs, {})
                           ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/origami/michelle/projects/boutiques/boutiques/invocationSchemaHandler.py", line 61, in addTypeConstraints
    h[id] = {"type": "array", "items": undertype()}
                                       ~~~~~~~~~^^
  File "/data/origami/michelle/projects/boutiques/boutiques/invocationSchemaHandler.py", line 58, in undertype
    return {"type": type.lower()}
                    ^^^^^^^^^^
AttributeError: 'collections.OrderedDict' object has no attribute 'lower'
```

PR #737, opened 2026-01-19
Ensure temporary files are cleaned up while running tests

### LINKED ISSUE #733: Ensure cleanup of temporary test files
## Context

Some temporary files from the test suite are not deleted properly

mathdugre:
## Related issues

fix #733

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

Prevent cluttering of local environment

## Current behaviour

Multiple tests leave a temp file. It is either due to the executor using (1) `--debug` flag or (2) testing expected failure with the executor raising exception  before cleanup.

## New behaviour

Temporary files are now cleaned up

- [ ] Yes
- [ ] No

### DIFF
--- a/boutiques/localExec.py
+++ b/boutiques/localExec.py
@@ -465,7 +465,7 @@ class LocalExecutor:
 
         # Destroy temporary docker script, if desired.
         # By default, keep the script so the dev can look at it.
-        if conIsPresent and not self.debug:
+        if ("pytest" in sys.modules) or (conIsPresent and not self.debug):
             if os.path.isfile(dsname):
                 os.remove(dsname)
 
--- a/boutiques/localExec.py
+++ b/boutiques/localExec.py
@@ -25,6 +25,14 @@ from boutiques.logger import print_info, print_warning, raise_error
 from boutiques.util.utils import conditionalExpFormat, extractFileName, loadJson
 
 
+def cleanup_temp_file(filename: str, *, debug: bool, conIsPresent: bool):
+    # Destroy temporary docker script, if desired.
+    # By default, keep the script so the dev can look at it.
+    if ("pytest" in sys.modules) or (conIsPresent and not debug):
+        if os.path.isfile(filename):
+            os.remove(filename)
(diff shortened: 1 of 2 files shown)


michellewang:
@mathdugre could you point to some test cases that create those temporary files?

michellewang:
@mathdugre when I run the tests locally I get a bunch of errors about `test_temp/test-data-cache`, do you know what that is about?

```
ERROR boutiques/tests/test_data_handler.py::TestDataHandler::test_publish_bulk - OSError: [Errno 16] Device or resource busy: 'boutiques/test_temp/test-data-cache'
```

Related: running the tests still creates this `test_temp` directory which is empty, do we want to address this in this PR as well?

boutiques/localExec.py
```diff
@@ -25,6 +25,14 @@
 from boutiques.util.utils import conditionalExpFormat, extractFileName, loadJson
 
 
+def cleanup_temp_file(filename: str, *, debug: bool, conIsPresent: bool):
+    # Destroy temporary docker script, if desired.
+    # By default, keep the script so the dev can look at it.
+    if ("pytest" in sys.modules) or (conIsPresent and not debug):
```
michellewang: Alternatively, we could create a Pytest "autouse" fixture that temporarily sets the working directory to be `tmp_path` while the test is running:

```python
# conftest.py
import os
from pathlib import Path

import pytest

@pytest.fixture(autouse=True)
def change_working_directory(tmp_path: Path):
    """Change working directory to a temporary path for the duration of the test."""
    original_cwd = Path.cwd()
    os.chdir(tmp_path)
    try:
        yield
    finally:
        os.chdir(original_cwd)
```

The benefit would be that, if the test fails, the temp file remains available and can be checked later.

@mathdugre what do you think?

mathdugre:
@michellewang I think I had this error in the past. I don't remember how to fix it, but I think I just restarted my computer.

I'll have a look for the `test_temp` folder and the fixture.

PR #745, opened 2026-01-20
Remove unused imports fix #744

### LINKED ISSUE #744: remove unused imports

mathdugre:
Re-enabled flak8 F401 and removed unused imports

- Closes #744

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -58,4 +58,4 @@ max_returns_amount = 30
 ;     # - docstrings rules that should not be applied to doc
 ; 	doc/*: D103, F401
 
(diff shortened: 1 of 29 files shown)


mathdugre:
fix #744

michellewang (commented):
@mathdugre could you explain the changes to the doc files? Specifically the addition of the three `.rst` files as it's not obvious to me why they were added to the PR

mathdugre:
I included those because the pre-commit hooks were complaining otherwise. I think the larger diff might be due to the file being generated automatically and not being ignore in the `.gitignore` file.
@michellewang 

I can spend more time trying to minimize the diff if needed.

Issue #694, opened 2024-11-12
DeprecationWarning importlib.load_module will be remove after 3.12


mathdugre:
DeprecationWarning: the [load_module()](https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.load_module) method is deprecated and slated for removal in Python 3.12; use [exec_module()](https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.exec_module) instead

Abhishek022001:
can i work on this issue?

mathdugre:
Yes, please go ahead

mathdugre:
Make use to use the [`develop`](https://github.com/boutiques/boutiques/tree/develop) branch as it contains the latest changes

Abhishek022001:
I am unable to locate the load_module() in the repo; can you pls help me?

mathdugre:
These are the place I found them, using global search in vscode
```python3
19 results - 8 files

.venv/boutiques/lib/python3.12/site-packages/_virtualenv.py:
  76                          is_new_api = hasattr(spec.loader, "exec_module")
  77:                         func_name = "exec_module" if is_new_api else "load_module"
  78                          old = getattr(spec.loader, func_name)
  79:                         func = self.exec_module if is_new_api else self.load_module
  80                          if old is not func:

  94      @staticmethod
  95:     def load_module(old, name):
  96          module = old(name)

.venv/boutiques/lib/python3.12/site-packages/six.py:
  204  
  205:     def load_module(self, fullname):
  206          try:

  236      def create_module(self, spec):
  237:         return self.load_module(spec.name)
  238  

  261      MovedAttribute("range", "__builtin__", "builtins", "xrange", "range"),
  262:     MovedAttribute("reload_module", "__builtin__", "importlib" if PY34 else "imp", "reload"),
  263      MovedAttribute("reduce", "__builtin__", "functools"),

.venv/boutiques/lib/python3.12/site-packages/pip/_vendor/six.py:
  204  
  205:     def load_module(self, fullname):
  206          try:

  236      def create_module(self, spec):
  237:         return self.load_module(spec.name)
  238  

  261      MovedAttribute("range", "__builtin__", "builtins", "xrange", "range"),
  262:     MovedAttribute("reload_module", "__builtin__", "importlib" if PY34 else "imp", "reload"),
  263      MovedAttribute("reduce", "__builtin__", "functools"),

[...]

PR #749, opened 2026-01-21
Update deprecated importlib load_module()

aptinis:
## Related issues
Closes issue #694

## Checklist

- [ ] **DO** Unit tests pass.

## Purpose

Update method to load docopt script as a module since `importlib.load_module` is deprecated and will be removed in Python 3.15.

## Current behaviour

`bosh import docopt` automatically generates a descriptors JSON file from a user provided Docopt script. Currently the user provided docopt script is loaded as a module using the deprecated `importlib.load_module()`.

## New behaviour

Functionality remains the same but PR loads the docopt script as a module using importlib recommended approach, see [example](https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly).

- [ ] Yes
- [x] No

## Implementation Detail

Intentionally defined a loader using `SourceFileLoader()` instead of using `spec_from_path()` because `spec_from_path()` checks for `.py` extension under the hood and will fail if the user provided docopt script has no extension.

### DIFF
--- a/boutiques/importer.py
+++ b/boutiques/importer.py
@@ -1,12 +1,10 @@
 #!/usr/bin/env python
-
 import collections
 import os
 import os.path as op
 import re
 import sys
 from argparse import ArgumentParser
-from importlib.machinery import SourceFileLoader
 
 import simplejson as json
 from jsonschema import ValidationError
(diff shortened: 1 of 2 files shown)


mathdugre:
Hi @aptinis , this looks great!!
We might want to add a test to ensure the functionality of the new `import_from_path` function. It could verify that:
1. A module is not currently imported
2. Import it with the function
3. Verify that the import worked

Let me know if you have questions.

